PR #18, opened 2018-07-18
check whether ip rule exists before add

cadmuxe:


pkg/config/config.go
```diff
@@ -100,9 +101,15 @@ func (r IPRouteConfig) Ensure(enabled bool) error {
 func (r IPRuleConfig) Ensure(enabled bool) error {
 	var err error
 	if enabled {
-		err = netlink.RuleAdd(&r.Rule)
-		if os.IsExist(err) {
-			err = nil
+		exist, e := r.exist()
```
jingax10: As we tested on CoS, we would run into a case to add the same rule multiple times. Can we change the logic to ensure the number of desired IP rule to be 1 when enabled is true while 0 when enabled is false?

I think the above logic is more robust as you have no assumption on what the set of IP rules looks like.

cadmuxe: yeah, I was debating on the same issue. will update.

cadmuxe: updated.

pkg/config/config.go
```diff
@@ -100,19 +101,53 @@ func (r IPRouteConfig) Ensure(enabled bool) error {
...
+				glog.Errorf("failed to delete duplicated ip rule: $v, error: %v", r.Rule, err)
+			}
+			ruleCount--
 		}
 	}
 
 	return err
 }
 
+func (r IPRuleConfig) exist() (int, error) {
```
jingax10: could we have a better name other than exist, e.g., ruleCount or count?

pkg/config/config.go
```diff
@@ -100,19 +101,53 @@ func (r IPRouteConfig) Ensure(enabled bool) error {
...
+	rules, err := netlink.RuleList(unix.AF_INET)
+	if err != nil {
+		return 0, err
+	}
+	var count int
```
jingax10: nit: count := 0

better to be explicitly initialized

pkg/config/config.go
```diff
@@ -100,19 +101,53 @@ func (r IPRouteConfig) Ensure(enabled bool) error {
 func (r IPRuleConfig) Ensure(enabled bool) error {
 	var err error
 	if enabled {
-		err = netlink.RuleAdd(&r.Rule)
-		if os.IsExist(err) {
-			err = nil
+		ruleCount, e := r.exist()
```
jingax10: can we have a helper function, say ensureRuleCount(int), then

if enabled {
  ensureRuleCount(1)
} else {
  ensureRuleCount(0)
}

cadmuxe:
Build failed with Go: master, succeed with 1.8, 1.9, 1.10. 
The failure due to golang

pkg/config/config.go
```diff
@@ -98,21 +99,50 @@ func (r IPRouteConfig) Ensure(enabled bool) error {
...
+	}
+
+	for ruleCount != ensureCount {
+		if ruleCount > ensureCount {
+			if err = netlink.RuleDel(&r.Rule); err != nil {
+				glog.Errorf("failed to delete duplicated ip rule: %v, error: %v", r.Rule, err)
+			}
+			ruleCount--
+		} else {
+			if err = netlink.RuleAdd(&r.Rule); os.IsExist(err) {
```
jingax10: what if the error is not IsExist, can you log an err?

pkg/config/config.go
```diff
@@ -98,21 +99,50 @@ func (r IPRouteConfig) Ensure(enabled bool) error {
...
+		return r.ensureHelper(1)
+	}
+	return r.ensureHelper(0)
+}
+
+func (r IPRuleConfig) ensureHelper(ensureCount int) error {
+	var err error
+	ruleCount, err := r.count()
+	if err != nil {
+		return err
```
jingax10: can you log an error?

jingax10:
@cadmuxe  Could you please squash your commits before merge?

PR #21, opened 2018-07-19
change internal struct/field to unexported.

cadmuxe:


jingax10:
koonwah@, you may need to think about this config package could be reused in other scenarios. So I would prefer to keep as it is. How do you think?

cadmuxe:
yeah, but if that happened, do we need to expose all the filed to other package?  If someday we really need to do that, we can add getter/export them again. 
Export them not hurt, but i think keep those structs internal before we have a reason is a good practice.

cadmuxe:
yeah, it make senses to me. I will close this pr.

PR #22, opened 2018-07-20
Add IPV6 subnet range and default route, if the node has an IPV6 interface

varunmar:
Tested by manually running on staging. No automated tests yet.

### DIFF
--- a/netd.yaml
+++ b/netd.yaml
@@ -75,12 +75,10 @@ data:
           "ipam": {
               "type": "host-local",
               "ranges": [
-                [
-                  {"subnet": podCidr}
-                ]
+              @ipv4Subnet@ipv6SubnetOptional
               ],
               "routes": [
-                {"dst": "0.0.0.0/0"}
+                {"dst": "0.0.0.0/0"}@ipv6RouteOptional
               ]
           }
         },
@@ -97,6 +95,7 @@ data:
   enable_policy_routing: "true"
   enable_masquerade: "true"
   enable_calico_network_policy: "false"
+  enable_private_ipv6_access: "false"
 
 ---
 
@@ -143,6 +142,11 @@ spec:
               configMapKeyRef:
                 name: netd-config
                 key: enable_calico_network_policy
+          - name: ENABLE_PRIVATE_IPV6_ACCESS
+            valueFrom:
+              configMapKeyRef:
+                name: netd-config
+                key: enable_private_ipv6_access
         volumeMounts:
         - mountPath: /host/etc/cni/net.d
           name: cni-net-dir
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -23,17 +23,42 @@ fi
 
 token=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
 node_url="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
-pod_cidr=$(curl -k -s -H "Authorization: Bearer $token" $node_url | jq '.spec.podCIDR')
-if [ -z "${pod_cidr:-}" ]; then
+ipv4_subnet=$(curl -k -s -H "Authorization: Bearer $token" $node_url | jq '.spec.podCIDR')
+if [ -z "${ipv4_subnet:-}" ]; then
   echo "Failed to fetch PodCIDR from K8s API server. Exiting with an error (1) ..."
   exit 1
 fi
 
 if [ -w /host/etc/cni/net.d ]; then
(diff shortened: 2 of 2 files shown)


scripts/install-cni.sh
```diff
@@ -27,8 +27,19 @@ fi
 if [ -w /host/etc/cni/net.d ]; then
   cni_spec=$(echo ${CNI_SPEC_TEMPLATE:-} | sed -e "s#podCidr#${pod_cidr:-}#g")
 
+  node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s" -H "Metadata-Flavor: Google" | jq '.[0]') ||:
```
jingax10: Should we fetch IPv6 conditioned on ENABLE_PRIVATE_IPV6_ACCESS?

It is possible the subnet is enabled for private IPv6 access but user does not enable private IPv6 access for the K8s cluster.

netd.yaml
```diff
@@ -77,10 +77,10 @@ data:
               "ranges": [
                 [
                   {"subnet": podCidr}
-                ]
+                ]ipv6RangeOptional
```
cadmuxe: Cloud we use some special character($#@ etc.) as the prefix, so that we can easily search/recognize the VARs?
Same for the podCidr.

varunmar: Good idea. I used @, since both # and $ look too much like shell characters.

scripts/install-cni.sh
varunmar: Done.

cadmuxe:
@varunmar Hi Varun, not sure what happened, but I can't see  your changes. only see Merge upstream commit, but looks like it doesn't contain your change.

varunmar:
Yes - sorry, I jumped the gun a bit. It should be there now.

netd.yaml
```diff
@@ -76,11 +76,11 @@ data:
               "type": "host-local",
               "ranges": [
                 [
-                  {"subnet": podCidr}
-                ]
+                  {"subnet": @podCidr}
```
jingax10: similar to what you introduced, maybe just replace [{"subnet": @podCidr}] as @ipv4Subnet

netd.yaml
```diff
@@ -76,11 +76,11 @@ data:
...
+                ]@ipv6RangeOptional
```
jingax10: nit: how about @ipv6SubnetOptional?

scripts/install-cni.sh
```diff
@@ -25,10 +25,25 @@ if [ -z "${pod_cidr:-}" ]; then
 fi
 
 if [ -w /host/etc/cni/net.d ]; then
-  cni_spec=$(echo ${CNI_SPEC_TEMPLATE:-} | sed -e "s#podCidr#${pod_cidr:-}#g")
```
jingax10: How about

1) figure out the values of both pid_cidr and node_ipv6_add;
2) generate the CNI spec once;
?

jingax10:
Varun, besides addressing the comments, let's have a manual verification first before merging your PR. I will update the steps for your reference.

scripts/install-cni.sh
jingax10: Also, please add logging properly to make debugging easier :-)

scripts/install-cni.sh
```diff
@@ -25,10 +25,25 @@ if [ -z "${pod_cidr:-}" ]; then
...
+  cni_spec=$(echo ${CNI_SPEC_TEMPLATE:-} | sed -e "s#@podCidr#${pod_cidr:-}#g")
```
jingax10: Also, as of now, we should not generate CNI spec if ENABLE_CALICO_NETWORK_POLICY is true.

scripts/install-cni.sh
```diff
@@ -25,10 +25,25 @@ if [ -z "${pod_cidr:-}" ]; then
...
+
+  if [ "$ENABLE_PRIVATE_IPV6_ACCESS" = true ]; then
+    node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ipv6s" -H "Metadata-Flavor: Google" | jq '.[0]') ||:
```
jingax10: I just tried the following commands, it works for me
"http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq '.ipv6s[0]'

BTW, what do -k and --fail mean for curl?

scripts/install-cni.sh
varunmar: I'd actually like to add the IPV4 subnet range separately from the IPV6 one, since the IPV6 one is optional, and the entire script should fail if we do not get the IPV4 range.

jingax10:
@varunmar changes look good to me. I just wonder whether you have successfully run with your changes or not.

Also, do not forget to squash your commits.

PR #24, opened 2018-07-24
add unit test

cadmuxe:


.travis.yml
```diff
@@ -11,4 +11,5 @@ install:
   - go get -u github.com/golang/dep/...
   - dep ensure
 script:
-  - make
\ No newline at end of file
+  - make
+  - make test
```
jingax10: nit: not sure about the last symbol, add a blank line by the end?

pkg/config/config.go
```diff
@@ -38,36 +37,58 @@ type Set struct {
...
+	NewChain(table, chain string) error
+	ClearChain(table, chain string) error
+	DeleteChain(table, chain string) error
+	AppendUnique(table, chain string, rulespec ...string) error
+	Delete(table, chain string, rulespec ...string) error
+}
+
 type IPTablesChainSpec struct {
 	TableName, ChainName string
 	IsDefaultChain       bool // Is a System default chain, if yes, we won't delete it.
```
jingax10: nit: predefined chain?

pkg/config/config.go
```diff
@@ -38,36 +37,58 @@ type Set struct {
...
+	Ipt                  IPTabler
 }
 
 type IPTablesRuleConfig struct {
 	Spec      IPTablesChainSpec
 	RuleSpecs []IPTablesRuleSpec
+	Ipt       IPTabler
```
jingax10: for golang style guide, for acroym, all should be capital, e.g., Html should be HTML

pkg/config/config_test.go
```diff
@@ -0,0 +1,218 @@
...
+	if count, _ := ipRule.count(); len(ruleList) != 1 || count != 0 {
+		t.Error("failed to delete IPRule")
+	}
+	ipRule.Ensure(true)
+	if count, _ := ipRule.count(); len(ruleList) != 2 || count != 1 {
+		t.Error("failed to delete IPRule")
+	}
+}
+
+type MockIPtable struct {
```
jingax10: nit: s/MockIPtable/MockIPTable/

.travis.yml
anfernee: +1, files normally should end with a newline.

pkg/config/config.go
```diff
@@ -38,36 +37,58 @@ type Set struct {
 	Configs     []Config
 }
 
+type Sysctler func(name string, params ...string) (string, error)
```
anfernee: exposed symbols should have comments, but it could be added later.

pkg/config/config_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	ipRule := IPRuleConfig{
+		Rule:     netlink.Rule{SuppressIfgroup: -1, SuppressPrefixlen: -1, Mark: -1, Mask: -1, Goto: -1},
+		RuleAdd:  func(rule *netlink.Rule) error { ruleList = append(ruleList, *rule); return nil },
+		RuleDel:  mockRuleDel,
+		RuleList: func(family int) ([]netlink.Rule, error) { return ruleList, nil },
+	}
+	if count, _ := ipRule.count(); count != 2 {
+		t.Errorf("IPRuleConfig.count() return wrong count.")
+	}
```
anfernee: this case covered by the case underneath?

pkg/config/config_test.go
anfernee: It's more like a fake than mock. the difference is subtle tho. mock only checks if a method is called and how many times it gets called. fake mimic the behavior of the interface. and fake normally requires a test for the fake itself, before it gets plugged in test.

pkg/config/config.go
cadmuxe: Yeah, I will add comments later.

PR #27, opened 2018-07-27
Fix install script bugs around handling ipv6 addresses

varunmar:
1. Remove the extra quotes around the IPV6 address returned from the Nanny
2. Handle the case where the address returned is "null" instead of an empty string
3. Add back the line continuation accidentally removed from a previous merge

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -34,13 +34,13 @@ if [ -w /host/etc/cni/net.d ]; then
   cni_spec=$(echo ${CNI_SPEC_TEMPLATE:-} | sed -e "s#@ipv4Subnet#[{\"subnet\": ${ipv4_subnet:-}}]#g")
 
   if [ "$ENABLE_PRIVATE_IPV6_ACCESS" = true ]; then
-    node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq '.ipv6s[0]') ||:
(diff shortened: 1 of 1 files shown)


varunmar:
Testing on devcluster is still in progress - PTAL in the meantime.

scripts/install-cni.sh
```diff
@@ -34,11 +34,11 @@ if [ -w /host/etc/cni/net.d ]; then
   cni_spec=$(echo ${CNI_SPEC_TEMPLATE:-} | sed -e "s#@ipv4Subnet#[{\"subnet\": ${ipv4_subnet:-}}]#g")
 
   if [ "$ENABLE_PRIVATE_IPV6_ACCESS" = true ]; then
-    node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq '.ipv6s[0]') ||:
+    node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq -r '.ipv6s[0]' ) ||:
```
cadmuxe: what does jq do ?

varunmar: jq is a json parser than Jing told me about :)
This change adds the '-r' (aka --raw) param which removes the quotes from the returned string.

cadmuxe: interesting, does our container contain this tool? I didn't see it in the build script.

varunmar: It's added in the dockerfile here: https://github.com/GoogleCloudPlatform/netd/blob/master/Dockerfile.in#L19

varunmar:
Ok, tested on devcluster and everything seems to up and running when enable_private_ipv6_access is true as well as when it is false.

jingax10:
@varunmar has verified it with the correctness of CNI spec.

/lgtm

PR #29, opened 2018-08-09
fix install-cni.sh

cadmuxe:


cadmuxe:
yeah, learned something new, but i will forget it soon :-)

PR #30, opened 2018-08-14
netd should log version info

cadmuxe:


scripts/install-cni.sh
```diff
@@ -15,6 +15,9 @@
 # limitations under the License.
 
 set -u -e
+
+echo "netd version: @VERSION@"
```
varunmar: This is fine, but FYI the other substitution variables below use a single @ sign before the name.

build/build.sh
```diff
@@ -31,6 +31,8 @@ if [ -z "${VERSION}" ]; then
     exit 1
 fi
 
+cat scripts/install-cni.sh | sed "s/@VERSION@/${VERSION}/" > scripts/install-cni-build.sh
```
varunmar: Suggest removing the unnecessary use of cat:
sed -e "s/@VERSION@/${VERSION}/g" scripts/install-cni.sh > scripts/install-cni-build.sh
Also consider if we want to surround the substitution string with @ or just at the beginning. I'm okay with either one, but let's be consistent.

cadmuxe: Oh, it got merged before I update it, will do it in the next pr.
Thanks.

PR #31, opened 2018-08-16
Add support for generating a Calico CNI Spec

varunmar:
This is only to generate the spec for calico, using the same spec name. Changes to the calico daemonset to use this file instead of generating its own are upcoming.

### DIFF
--- a/netd.yaml
+++ b/netd.yaml
@@ -91,6 +91,43 @@ data:
         }
       ]
     }
+  calico_cni_spec_template: |-
+    {
+      "name": "k8s-pod-network",
+      "cniVersion": "0.3.0",
+      "plugins": [
+        {
+        "type": "calico",
+        "log_level": "debug",
+        "datastore_type": "kubernetes",
+        "nodename": "__KUBERNETES_NODE_NAME__",
+        "ipam": {
+          "type": "host-local",
+          "ranges": [
+          [ { "subnet": "usePodCidr" } ]@ipv6SubnetOptional
+          ],
+          "routes": [
+            {"dst": "0.0.0.0/0"}@ipv6RouteOptional
+          ]
+        },
+        "policy": {
+          "type": "k8s",
+          "k8s_auth_token": "__SERVICEACCOUNT_TOKEN__"
+        },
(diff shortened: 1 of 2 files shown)


netd.yaml
```diff
@@ -91,6 +91,43 @@ data:
         }
       ]
     }
+  cni_calico_spec_template: |-
```
jingax10: nit: s/cni_calico_spec_template/calico_cni_spec_template/

netd.yaml
```diff
@@ -132,6 +169,11 @@ spec:
               configMapKeyRef:
                 name: netd-config
                 key: cni_spec_template
+          - name: CNI_CALICO_SPEC_TEMPLATE
```
jingax10: nit: s/CNI_CALICO_SPEC_TEMPLATE/CALICO_CNI_SPEC_TEMPLATE/

scripts/install-cni.sh
```diff
@@ -16,8 +16,10 @@
 
 set -u -e
 if [ "${ENABLE_CALICO_NETWORK_POLICY}" == "true" ]; then
-  echo "Calico Network Policy is enabled by ENABLE_CALICO_NETWORK_POLICY. Disabling CNI Spec generation."
-  exit 0
+  echo "Calico Network Policy is enabled by ENABLE_CALICO_NETWORK_POLICY. Generating Calico spec."
```
jingax10: nit: s/Generating Calico spec/Generating Calico CNI spec/

netd.yaml
varunmar: Done.

netd.yaml
varunmar: Done.

scripts/install-cni.sh
varunmar: Done.

varunmar:
Tested with manual fixes to calico-node-daemonset.yaml
The CNI spec template works, so I'm going to merge this in.

jingax10:
/lgtm

@varunmar, thanks for verification! Please also port your change to custom calico yaml change for GKE.

PR #34, opened 2018-09-13
Update netd DaemonSet's updateStrategy to be RollingUpdate and tolerations.

jingax10:
We need to update it for GKE and testing as well.

### DIFF
--- a/netd.yaml
+++ b/netd.yaml
@@ -148,6 +148,10 @@ spec:
   selector:
     matchLabels:
       k8s-app: netd
+  updateStrategy:
+    type: RollingUpdate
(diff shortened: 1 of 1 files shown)


netd.yaml
```diff
@@ -158,6 +162,11 @@ spec:
       terminationGracePeriodSeconds: 0
       nodeSelector:
         beta.kubernetes.io/kube-netd-ready: "true"
+      tolerations:
```
cadmuxe: The tolerations will match all the nodes, is this your purpose? 
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

Issue #32, opened 2018-09-12
Make config reconcile interval configurable


jingax10:
Right now, the config reconcile interval is hard coded as 10s in https://github.com/GoogleCloudPlatform/netd/blob/master/pkg/controllers/netconf/network_config_controller.go#L58. We should make it configurable via ConfigMap.

cadmuxe:
Done with https://github.com/GoogleCloudPlatform/netd/pull/35

PR #35, opened 2018-09-13
Make reconcile interval configurable.

cadmuxe:
https://github.com/GoogleCloudPlatform/netd/issues/32


netd.yaml
```diff
@@ -133,6 +133,7 @@ data:
   enable_masquerade: "true"
   enable_calico_network_policy: "false"
   enable_private_ipv6_access: "false"
+  reconcile_interval_seconds: 10
```
jingax10: nit: how about 30s or 60s?

PR #41, opened 2018-09-19
ensure network configs immediately after enter the Run loop.

cadmuxe:
Fix yaml


jingax10:
/lgtm

p.s. It is important to test against every PR :-)

cadmuxe:
Yeah, the main purpose of this pr is to run ensure before the first loop. otherwise the create cluster will need an extra 60s for the current config. I have tested the binary directly.

PR #42, opened 2018-09-24
change nodeSelector label to gke instead of k8s, update the addonmanager

cadmuxe:
mode to Reconcile

### DIFF
--- a/netd.yaml
+++ b/netd.yaml
@@ -19,7 +19,7 @@ metadata:
   namespace: kube-system
   labels:
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
 rules:
 - apiGroups: [""]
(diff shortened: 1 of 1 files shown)


netd.yaml
```diff
@@ -162,7 +162,7 @@ spec:
       serviceAccountName: netd
       terminationGracePeriodSeconds: 0
       nodeSelector:
-        beta.kubernetes.io/kube-netd-ready: "true"
+        cloud.google.com/gke-netd-ready: "true"
```
varunmar: I thought thockin@ said something about gke.io instead of using cloud.google.com?

cadmuxe: node_startup.go is using cloud.google.com

jingax10:
/lgtm

Please test it out in GKE before submit, e.g., change the ConfigMap, DaemonSet configs to see whether it could be reconciled. Thanks!

cadmuxe:
tested with cl/214349135

Issue #43, opened 2018-09-28
PR #31 shouldn't generate the template at /etc/cni/net.d/


cadmuxe:
https://github.com/GoogleCloudPlatform/netd/pull/31
Calico is not looking for the existing .conflist.
maybe the config should go to g3/calico-node-daemonset-template.yaml.in instead?

varunmar:
We set the CNI_NETWORK_CONFIG_FILE to the same .conflist file as netd. That is wrong, and we should move it somewhere else.

PR #44, opened 2018-09-29
Quick fix to avoid unnecessary CNI spec generation.

jingax10:
Instead of rollback #31 completely, here is a temp & quick fix for #43. @varunmar will fix in a follow-up PR.


cadmuxe:
I have tested the fix, working good.
/lgtm

PR #50, opened 2018-10-25
 Add network metrics collection

lzang:
Add the network metrics collection logic to collect the following metrics:
- conntrack entries: conntrack usage
- conntrack_error_count (insert_failed, and drop) : A request from TSE due to the issue mentioned here 
     https://www.weave.works/blog/racy-conntrack-and-dns-lookup-timeouts
- num_inuse_sockets(tcp, udp): this would help to tell if the node is under some sort of ddos attack, or a particular node has an unusually high number of connections.
- num_tw_sockets: this can tell if there is a large number of short-lived sockets, which is often not desired.
- socket_memory: the memory used by all sockets. When the memory hits the limit, the network becomes not usable. There is a request to add this metrics (https://github.com/kubernetes/kubernetes/issues/62334).
All these metrics can be obtained by read proc file directly (i.e. O(1) cost) .


pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+// NodeCollector implements the prometheus.Collector interface.
+type nodeCollector struct {
+	Collectors map[string]Collector
+}
+
+func init() {
+	startTime = time.Now().Unix()
+}
+
+// NewNodeCollector creates a new NodeCollector
```
bowei: you need to document `filters` as it is string type

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+func NewNodeCollector(filters []string, proc string) (*nodeCollector, error) {
+	procPath = proc
+	collectors := make(map[string]Collector)
+	for _, filter := range filters {
+		createFunc, exist := factories[filter]
+		if !exist {
+			return nil, fmt.Errorf("missing collector: %s", filter)
```
bowei: use `%q`

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+		} else {
+			if c, err := createFunc(); err == nil {
+				collectors[filter] = c
+			} else {
+				return nil, fmt.Errorf("fail to create collector: %s, error %v", filter, err)
```
bowei: it's better to return err directly, than wrap it
```go
c, err := createFunc()
if err != nil {
  return nil, err
}
collectors[filter] = c
```

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+		nil,
+	)
+	scrapeSuccessDesc = prometheus.NewDesc(
+		"scrape_collector_success",
+		"Whether a collector succeeded.",
+		[]string{"collector"},
+		nil,
+	)
+	// Needed by prometheus-to-sd
+	scrapeStartDesc = prometheus.NewDesc(
```
bowei: this should be called processStart... not scrape start

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+	var success float64
+
+	if err != nil {
+		glog.Infof("ERROR: %s collector failed after %fs: %s", name, duration.Seconds(), err)
+		success = 0
+	} else {
+		success = 1
+	}
+	ch <- prometheus.MustNewConstMetric(scrapeDurationDesc, prometheus.GaugeValue, duration.Seconds(), name)
+	ch <- prometheus.MustNewConstMetric(scrapeSuccessDesc, prometheus.GaugeValue, success, name)
```
bowei: This seems wrong -- this should be a cumulative value

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+			wg.Done()
+		}(name, c)
+	}
+	wg.Wait()
+}
+
+func execute(name string, c Collector, ch chan<- prometheus.Metric) {
+	begin := time.Now()
+	err := c.Update(ch)
+	duration := time.Since(begin)
```
bowei: Can we not use `NewTimer.ObserveDuration()`

https://github.com/prometheus/client_golang/blob/master/prometheus/timer.go#L33

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+	var success float64
+
+	if err != nil {
+		glog.Infof("ERROR: %s collector failed after %fs: %s", name, duration.Seconds(), err)
+		success = 0
+	} else {
+		success = 1
+	}
+	ch <- prometheus.MustNewConstMetric(scrapeDurationDesc, prometheus.GaugeValue, duration.Seconds(), name)
```
bowei: This should be histogram?

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,42 @@
...
+*/
+
+package collector
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestConntrack(t *testing.T) {
```
bowei: More unit tests needed:

* file is not there
* error reading file
* invalid entries
* bad parse
* zero length file

pkg/metrics/collector/helper.go
```diff
@@ -0,0 +1,40 @@
...
+package collector
+
+import (
+	"io/ioutil"
+	"path"
+	"strconv"
+	"strings"
+)
+
+func readUintFromFile(path string) (uint64, error) {
```
bowei: docstring

pkg/metrics/collector/helper.go
```diff
@@ -0,0 +1,40 @@
...
+		return 0, err
+	}
+	value, err := strconv.ParseUint(strings.TrimSpace(string(data)), 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	return value, nil
+}
+
+func procFilePath(name string) string {
```
bowei: docstring

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type sockStatCollector struct {
+	inuse  *prometheus.Desc
+	tw     *prometheus.Desc
+	memory *prometheus.Desc
+}
+
+var pageSize = os.Getpagesize()
```
bowei: why cache this value?

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,169 @@
...
+		memory: prometheus.NewDesc(
+			"socket_memory",
+			"Amount of memory used by sockets",
+			nil, nil,
+		),
+	}, nil
+}
+
+func initSocketStats() map[string]uint64 {
+	stats := map[string]uint64{
```
bowei: why use a map with strings instead of just a struct?
```
type stats struct {
  tcpInUse uint64
  ...
}
```

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,169 @@
...
+		"tcp_inuse": 0,
+		"udp_inuse": 0,
+		"tcp_tw":    0,
+		"mem":       0,
+	}
+	return stats
+}
+
+func getStatsName(protocol, name string) string {
```
bowei: this function looks like it will result in tech debt as it assumes a naming schem

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,169 @@
...
+	return parseSockStats(file)
+}
+
+func parseSockStats(r io.Reader) (map[string]map[string]string, error) {
+	sockStat := make(map[string]map[string]string)
+	scanner := bufio.NewScanner(r)
+
+	for scanner.Scan() {
+		line := strings.Split(scanner.Text(), " ")
+		// Remove trailing ':'.
```
bowei: shouldn't we check that it has a trailing ':' before blindly chopping it off?

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,169 @@
...
+		protocol := line[0][:len(line[0])-1]
```
bowei: we need to check that the line is non-empty

pkg/metrics/collector/sockstat_test.go
```diff
@@ -0,0 +1,62 @@
...
+limitations under the License.
+*/
+
+package collector
+
+import (
+	"strings"
+	"testing"
+)
+
```
bowei: need test cases with invalid input at least

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	listenAddress string
+	procPath      string
+}
+
+var mcfg metricsConfig
+
+func init() {
+	flag.StringVar(&mcfg.filter, "metrics-collectors", "", "Enable given metrics collectors (options: conntrack,socket).")
+	flag.StringVar(&mcfg.listenAddress, "metrics-port", ":8000", "Port on which to expose metrics.")
+	flag.StringVar(&mcfg.procPath, "metrics-proc-path", "/proc", "Proc directory to read metrics.")
```
bowei: each proc file read probably should be an individual flag.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	filter        string
+	listenAddress string
+	procPath      string
+}
+
+var mcfg metricsConfig
+
+func init() {
+	flag.StringVar(&mcfg.filter, "metrics-collectors", "", "Enable given metrics collectors (options: conntrack,socket).")
+	flag.StringVar(&mcfg.listenAddress, "metrics-port", ":8000", "Port on which to expose metrics.")
```
bowei: scrape interval?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+type metricsConfig struct {
+	filter        string
+	listenAddress string
+	procPath      string
+}
+
+var mcfg metricsConfig
+
+func init() {
+	flag.StringVar(&mcfg.filter, "metrics-collectors", "", "Enable given metrics collectors (options: conntrack,socket).")
```
bowei: this seems better named: "collectors" or "enabled-collectors"
this is not really a filter, more like a list of enable collectors

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+		registry,
+		promhttp.HandlerFor(gatherers,
+			promhttp.HandlerOpts{
+				ErrorHandling: promhttp.ContinueOnError,
+			}),
+	)
+	h.ServeHTTP(w, r)
+}
+
+func StartCollector() error {
```
bowei: docstring

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	if mcfg.filter == "" {
+		glog.Infof("Disable metrics collectors.")
```
bowei: "No metrics collectors were enabled"

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	flag.StringVar(&mcfg.listenAddress, "metrics-port", ":8000", "Port on which to expose metrics.")
+	flag.StringVar(&mcfg.procPath, "metrics-proc-path", "/proc", "Proc directory to read metrics.")
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	gatherers := prometheus.Gatherers{
+		registry,
+	}
+	// Delegate http serving to Prometheus client library, which will call collector.Collect.
+	h := promhttp.InstrumentMetricHandler(
```
bowei: Does it make sense to construct this repeatedly in the handler() routine for every HTTP get()

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+		nil,
+	)
+	startTime int64
+)
+
+var (
+	factories = make(map[string]func() (Collector, error))
+)
+
+func registerCollector(collector string, factory func() (Collector, error)) {
```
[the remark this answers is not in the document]
lzang: Added.

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+			} else {
+				return nil, fmt.Errorf("fail to create collector: %s, error %v", filter, err)
+			}
+		}
+	}
+	return &nodeCollector{Collectors: collectors}, nil
+}
+
+// Describe implements the prometheus.Collector interface.
+func (n nodeCollector) Describe(ch chan<- *prometheus.Desc) {
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+		"Duration of a collector scrape.",
+		[]string{"collector"},
+		nil,
+	)
+	scrapeSuccessDesc = prometheus.NewDesc(
+		"scrape_collector_success",
+		"Whether a collector succeeded.",
+		[]string{"collector"},
+		nil,
+	)
```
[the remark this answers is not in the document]
lzang: This just tells if the current call for a particular collector is successful or not, not really for stats purpose. This code was ported from nod exporter, and it is mainly used for self debugging.

pkg/metrics/collector/collector.go
lzang: Done

pkg/metrics/collector/collector.go
lzang: Updated

pkg/metrics/collector/collector.go
lzang: Updated.

pkg/metrics/collector/collector.go
lzang: I don't get it - do you mean we should use NewTimer.ObserveDuration()? as currently we didn't use it.

pkg/metrics/collector/collector.go
lzang: This just signal the duration for current call, and for self diagnose purpose, not stats across multiple samples. Looks different from histogram definition "A histogram samples observations (usually things like request durations or response sizes) and counts them in configurable buckets. It also provides a sum of all observed values" from https://prometheus.io/docs/concepts/metric_types/.

pkg/metrics/collector/conntrack_test.go
lzang: Updated the test.

pkg/metrics/collector/helper.go
lzang: Updated.

pkg/metrics/collector/collector.go
bowei: we should try to use standard library constructs if possible...

pkg/metrics/collector/helper.go
lzang: Updated.

pkg/metrics/collector/sockstat.go
lzang: Because the number we read from the proc file is the number of pages. To get the actual memory usage, we need to multiply the number to the page size.

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Describe implements the prometheus.Collector interface.
+func (n nodeCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- scrapeStartDesc
+	ch <- scrapeDurationDesc
+	ch <- scrapeSuccessDesc
+}
+
+// Collect implements the prometheus.Collector interface.
+func (n nodeCollector) Collect(ch chan<- prometheus.Metric) {
```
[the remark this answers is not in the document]
lzang: updated.

pkg/metrics/collector/collector.go
bowei: For recording latency of processing something, we typically use histogram. What makes this case different?

pkg/metrics/collector/collector.go
lzang: updated.

pkg/metrics/collector/collector.go
lzang: It is the data for current run. So not a cumulative value.

pkg/metrics/collector/sockstat.go
lzang: Because map is easier for automatic data assignment and integration.

pkg/metrics/collector/sockstat.go
lzang: This function is used to parse the data from proc file, so it does assume the format of the file and will convert the name correspondingly. The function is used locally.

pkg/metrics/collector/sockstat.go
lzang: Added more checks.

pkg/metrics/collector/sockstat.go
lzang: Added more checks.

pkg/metrics/collector/sockstat_test.go
lzang: Added test case.

pkg/metrics/metrics.go
lzang: All stat files used right now has only one common proc directory, so one shall be good.

pkg/metrics/metrics.go
lzang: This is decided by the polling agent prometheus-to-sd, not by collector.

pkg/metrics/metrics.go
lzang: Updated.

pkg/metrics/metrics.go
lzang: Updated.

pkg/metrics/metrics.go
lzang: Updated.

pkg/metrics/metrics.go
lzang: Updated.

pkg/metrics/collector/collector.go
lzang: The code is doing the same with that in the function. Current way allows us to print the duration in the error message followed.

pkg/metrics/collector/helper.go
bowei: you need to follow the go style comment. Make sure your code passes golint and govet

// readUintFromFile reads a uint ...

pkg/metrics/collector/collector.go
lzang: Update the diff to create histogram for this measure.

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,65 @@
...
+	if v, ok := counters["insert_failed"]; !ok || v != 24 {
+		t.Errorf("expect insert_failed 24, got %v", v)
+	}
+	if v, ok := counters["drop"]; !ok || v != 37 {
+		t.Errorf("expect drop 37, got %v", v)
+	}
+}
+
+func TestConntrackInvalidInput(t *testing.T) {
+	inputs := []string{
```
bowei: `t.Parallel()`

use the standard golang testing idiom:

```go
for _, tc := range []struct{
  desc string
  input string
} {
  {
     desc: "empty file",
     input: "",
   },
   ...
} {
   t.Run(tc.desc, func(t testing.T) {
     _, err := parseConntrackCounters(strings.NewReader(tc.input)
    if err == nil {
      t.Errorf("parseConntrackCounters(%q) = _, nil, want error", tc.input)
    }
}
```

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+package collector
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestConntrack(t *testing.T) {
+	input := "entries  searched found new invalid ignore delete delete_list insert insert_failed drop early_drop icmp_error  expect_new expect_create expect_delete search_restart\n" +
```
bowei: see below, structure your test using the standard golang testing idioms

bowei: what happens if the column did not exist?

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,165 @@
...
+						return nil, fmt.Errorf("invalid value %q in sockstats: %v", line[i+1], err)
+					} else {
+						ss.tcpTw = v
+					}
+				case "mem":
+					v, err := strconv.ParseUint(line[i+1], 10, 64)
+					if err != nil {
+						return nil, fmt.Errorf("invalid value %q in sockstats: %v", line[i+1], err)
+					} else {
+						ss.mem += v
```
bowei: why plus here vs set in the other places?

pkg/metrics/collector/sockstat_test.go
```diff
@@ -0,0 +1,76 @@
...
+		"TCP: inuse 23 orphan 0 tw 10 alloc 52 mem 6\n" +
+		"UDP: inuse 3 mem 3\n" +
+		"UDPLITE: inuse 0\n" +
+		"RAW: inuse 0\n" +
+		"FRAG: inuse 0 memory 0"
+	input6 := "TCP6: inuse 14\n" +
+		"UDP6: inuse 1\n" +
+		"UDPLITE6: inuse 0\n" +
+		"RAW6: inuse 0\n" +
+		"FRAG6: inuse 0 memory 0"
```
bowei: for mem, don't we also need to add in the RAW and FRAG memory?

pkg/metrics/collector/sockstat_test.go
```diff
@@ -0,0 +1,76 @@
...
+	if ss.tcpTw != 10 {
+		t.Errorf("expect tw 10, got %v", ss.tcpTw)
+	}
+	if ss.mem != 9 {
+		t.Errorf("expect mem 9, got %v", ss.mem)
+	}
+}
+
+func TestSockstatsInvalidInput(t *testing.T) {
+	inputs := []string{
```
bowei: see above, use the golang testing idioms

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/netd/pkg/metrics/collector"
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+type metricsConfig struct {
```
bowei: if this type is not used anywhere, you can use anonymous type:

```
var mcfg struct {
   enabledCollectors string
   ...
}
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+
+var mcfg metricsConfig
+
+func init() {
+	flag.StringVar(&mcfg.enabledCollectors, "metrics-collectors", "", "Enable given metrics collectors (options: conntrack,socket).")
+	flag.StringVar(&mcfg.listenAddress, "metrics-port", ":8000", "Port on which to expose metrics.")
+	flag.StringVar(&mcfg.procPath, "metrics-proc-path", "/proc", "Proc directory to read metrics.")
+}
+
+func StartCollector() error {
```
bowei: missing doc string

bowei:
Seems like my comment got eaten:
```
type conntrackStats struct {
  insertFailed int
  drop int
}

func (c *conntrackStats) merge(other *conntrackStats) {
  c.insertFailed += other.insertFailed
  c.drop += other.drop
}

type conntrackIndices struct {
  numFields int
  insert_failed_index int
  drop_index int
}

// parseHeader parses the conntrack header line, returning the
// indexes of the fields we wish to extract.
func parseHeader(line string) (conntrackIndicies, error) {
  var indicies conntrackIndicies
  nameParts = strings.Split(scanner.Text(), " ")
  indicies.numFields = len(nameParts)
  for i, v := range nameParts {
    switch v {
      case "insert_failed":
        indicies.insertFailed = i
      case "drop":
        indicies.drop = i
    }
  }
}

// parseConntrackData parses ...
func parseConntrackData(line string, indicies conntrackIndicies) (*conntrackStats, error) {
  stats := &conntrackStats{}
  valueParts := strings.Split(scanner.Text(), " ")
  if len(valueParts) != numFields { ... }

  if v, err := strconv.ParseUint(valueParts[indicies.insertFailed], 16, 32); err == nil {
    stats.insertFailed = v
  } else { ... }

  if v, err := strconv.ParseUint(valueParts[indicies.drops], 16, 32); err == nil {
    stats.drops = v
  } else { ... }

  return stats, nil
}

// parseConntrackFile parses the conntrack file contents read in from `r` and returns a merged set of stats.
func parseConntrackFile(r io.Reader) (conntrackStats, error) {
  ...
  if ! scanner.Scan() {
    // error no header line
  }
  indicies, err := parseHeader(scanner.Line())
  if err != nil { ... } 
  
  var accStats conntrackStats
  for scanner.Scan() {
     stats, err := parseConntrackData(scanner.Line(), indicies)
     if err != nil { ... }
     accStats.merge(stats)
  }
  ...
}
```

cmd/netd/main.go
```diff
@@ -50,6 +51,11 @@ func main() {
 	glog.Infof("Starting netd")
 	go nc.Run(stopCh, &wg)
 
+	err := metrics.StartCollector()
+	if err != nil {
+		glog.Infof("Fail to start metrics collector: %v!", err)
```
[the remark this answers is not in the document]
lzang: Didn't want to crash the main program for a non-critical feature. Agree though for the right input this failure should not happen, so I can change it to fatal.

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var procPath string
```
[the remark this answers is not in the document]
lzang: This one is actually used in function procFilePath() in helper.go. I can move the variable to the file

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,130 @@
...
+	"io"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type conntrackCollector struct {
+	current  *prometheus.Desc
```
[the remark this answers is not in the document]
lzang: current is the current number of conntrack entries. It is updated as reading a number from the file 
/proc/sys/net/netfilter/nf_conntrack_count. The other error counters need a little more parsing. 
Will update the name.

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func execute(name string, c Collector, ch chan<- prometheus.Metric) {
+	begin := time.Now()
+	err := c.Update(ch)
+	duration := time.Since(begin)
+
+	scrapeDurationHist.WithLabelValues(name).Observe(duration.Seconds())
+	if err != nil {
+		glog.Infof("ERROR: %s collector failed after %fs: %s", name, duration.Seconds(), err)
```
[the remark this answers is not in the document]
lzang: is this log level runtime adjustable? If a collector error happens, this print can flood the log as the collector func is called periodically.

pkg/metrics/collector/sockstat.go
lzang: Because this mem is the sum of tcp mem and udp mem. The others are not sum across tcp and udp.

pkg/metrics/collector/sockstat_test.go
lzang: RAW doesn't have mem stats. FRAG memory is using a different unit with tcp/udp mem, and probably not common in GKE setup as well.

cmd/netd/main.go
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/collector.go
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,129 @@
...
+		"process_start_time_seconds",
+		"Start time of the process since unix epoch in seconds.",
+		nil,
+		nil,
+	)
+	startTime int64
+)
+
+var (
+	factories = make(map[string]func() (Collector, error))
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,129 @@
...
+	factories[collector] = factory
+}
+
+// nodeCollector implements the prometheus.Collector interface.
+type nodeCollector struct {
+	Collectors map[string]Collector
+}
+
+func init() {
+	startTime = time.Now().Unix()
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/collector.go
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,130 @@
...
+				return nil, fmt.Errorf("cannot find wanted conntrack metrics")
+			}
+			numFields = len(nameParts)
+		} else {
+			valueParts := strings.Split(scanner.Text(), " ")
+			if len(valueParts) != numFields {
+				return nil, fmt.Errorf("Line %d doesn't contain enough fields", line)
+			}
+			for _, i := range wantedIndex {
+				if s, err := strconv.ParseUint(valueParts[i], 16, 32); err == nil {
```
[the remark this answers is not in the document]
lzang: Updated

pkg/metrics/collector/conntrack_test.go
lzang: Will report error

pkg/metrics/collector/conntrack_test.go
lzang: Updated

pkg/metrics/collector/sockstat.go
lzang: Added the comment.

pkg/metrics/collector/sockstat_test.go
lzang: Unchanged due to reason above.

pkg/metrics/collector/sockstat_test.go
lzang: Updated

pkg/metrics/metrics.go
lzang: Updated

pkg/metrics/metrics.go
lzang: Added

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,129 @@
...
+	processStartDesc = prometheus.NewDesc(
+		"process_start_time_seconds",
+		"Start time of the process since unix epoch in seconds.",
+		nil,
+		nil,
+	)
+	startTime int64
+)
+
+var (
```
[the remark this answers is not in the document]
lzang: Updated

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,86 @@
...
+			t.Parallel()
+			_, err := parseConntrackFile(strings.NewReader(tc.input))
+			if err == nil {
+				t.Errorf("parseConntrackFile(%q) = _, nil, want error", tc.input)
+			}
+		})
+	}
+
+	if _, err := getConntrackStats("./fake"); err == nil {
+		t.Errorf("getConntrackStats for a fake file returns nil, want error")
```
bowei:
```
t.Errorf(`getConntrackStats("./fake") = %v; want nil`, err)
```
(gostyle)

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+	for scanner.Scan() {
+		lineCount++
+		in := scanner.Text()
+		line := strings.Split(in, " ")
+		if len(line) < 2 || len(line[0]) < 2 || line[0][len(line[0])-1] != ':' {
+			return nil, fmt.Errorf("invalid input %q", in)
+		}
+		// Remove trailing ':'.
+		protocol := line[0][:len(line[0])-1]
+		if protocol == "TCP" || protocol == "TCP6" {
```
bowei: let factor this into several functions for easier maintenance:

```go
func (s*socketStats) merge(o *socketStats) {
  s.tcpInUse = o.tcpInUse
  ...
  // Sum memory used across protocols.
  s.mem += o.mem
}

func extractSockStatUint(field string, s string, out interface{}) error {
  v, err := strconv.ParseUint(s, 10, 64)
  if err != nil {
    return nil, fmt.Errorf("invalid value %q for field %q in sockstats", s, field)
  }
  outValue = out.(*uint64)
  *outValue = v
  return nil
}

type sockStatEntry struct {
  out interface{}
  extract func(string, string, interface{}) error
}

func extractSockStatFields(fields []string, entries map[string]sockStatEntry) error {
  for i := 0; i < len(fields) && i+1 < len(fields); i += 2 {
    k, v := fields[i], fields[i+1]
    if e, ok := entries[k]; !ok {
      continue
    }
    if err := e.extract(k, v, e.out); err != nil {
      return err
    } 
  }
  return nil
}

func parseLine(in string) (*socketStats, error) {
  line := strings.Split(in, " ")
  if len(line) < 2 || len(line[0]) < 2 || line[0][len(line[0])-1] != ':' {
    return nil, fmt.Errorf("invalid line %q", in)
  }

  tcpEntries := map[string]*sockStatEntry{
      "inuse": {&ret.tcpInUse, extractSockStatUint},
      "tw": {&ret.tcpTw, extractSockStatUint},
      "mem": {&ret.mem, extractSockStatUint},
  }

  udpEntries := map[string]*sockStatEntry{
    "inuse": {&ret.tcpInUse, extractSockStatUint},
    "tw": {&ret.tcpTw, extractSockStatUint},
  }

  ret := &socketStats{}
  protocol := line[0][:len(line[0])-1]
  rest := line[1:]

  switch protocol {
  case "TCP", "TCP6":
    if err := extractSockStatFields(rest, tcpEntries); err != nil {
      return err
    }
  case "UDP", "UDP6":
    if err := extractSockStatFields(rest, udpEntries); err != nil {
      return err
    }
  }
  return nil
}

func parseSockStats(r io.Reader) (*socketStats, error) {
  ...
  ret := &socketStats{}
  for scanner.Scan() {
    lineCount++
    if s, err := parseLine(scanner.Text()); err == nil {
      ret.merge(s)
    } else {
      return nil, err
    }
  }
  ...
}
```

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+	if lineCount == 0 {
+		return nil, fmt.Errorf("Empty file")
+	}
+	return &ss, nil
+}
+
+func (c *sockStatCollector) Update(ch chan<- prometheus.Metric) error {
+
```
bowei: extra line

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+	ss, err := getSockStats(procFilePath("net/sockstat"))
+	if err != nil {
+		return fmt.Errorf("couldn't get sockstats: %v", err)
```
bowei: preference is to not wrap errors unless there is a good reason
`return err`

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	ss6, err := getSockStats(procFilePath("net/sockstat6"))
+	if err != nil {
```
bowei: preference is to not wrap errors unless there is a good reason
`return err`

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+		return fmt.Errorf("couldn't get sockstats6: %v", err)
+	}
+
+	ch <- prometheus.MustNewConstMetric(
```
bowei: one line (gostyle)

fix other locations

pkg/metrics/collector/sockstat_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+package collector
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestSockstats(t *testing.T) {
+	input := "sockets: used 270\n" +
```
bowei: const

pkg/metrics/collector/sockstat_test.go
```diff
@@ -0,0 +1,97 @@
...
+		"TCP: inuse 23 orphan 0 tw 10 alloc 52 mem 6\n" +
+		"UDP: inuse 3 mem 3\n" +
+		"UDPLITE: inuse 0\n" +
+		"RAW: inuse 0\n" +
+		"FRAG: inuse 0 memory 0"
+	input6 := "TCP6: inuse 14\n" +
```
bowei: const

pkg/metrics/collector/sockstat_test.go
bowei: use the literal quote
```
const input = `sockets: used 270
TCP: inuse 23 orphan 0 tw 10 alloc 52 mem 6
...
`
```

pkg/metrics/collector/sockstat_test.go
bowei: use literal quote

pkg/metrics/collector/sockstat_test.go
```diff
@@ -0,0 +1,97 @@
...
+		"UDP6: inuse 1\n" +
+		"UDPLITE6: inuse 0\n" +
+		"RAW6: inuse 0\n" +
+		"FRAG6: inuse 0 memory 0"
+
+	ss, err := parseSockStats(strings.NewReader(input))
+	if err != nil {
+		t.Errorf("parseSockStats returns err %v, want nil", err)
```
bowei: t.Errorf("parseSockStats(%q) = _, %v; want _, nil", err)

pkg/metrics/collector/sockstat_test.go
bowei: t.Parallel()

bowei: you have multiple tests cases. Use the t.Run in the standard go test idiom:
```go
for _, tc := range []struct{
  desc string
  in string
  want socketStats   
}{
  {
    desc: "ipv4",
    ...
  },
  {
    desc: "ipv6",
    ...
  },
}{
  t.Run(tc.desc, func (t *testing.T) {
    ss, err := parseSocketStats(...)
    if err != nil {
      t.Errorf("parseSocketStats(%q) = _, %v; want _, nil", tc.in, err)
    }
    ...
  })
}


```

pkg/metrics/collector/sockstat_test.go
```diff
@@ -0,0 +1,97 @@
...
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+			_, err := parseSockStats(strings.NewReader(tc.input))
+			if err == nil {
+				t.Errorf("parseSockStats(%q) = _, nil, want error", tc.input)
+			}
+		})
+	}
+
+	if _, err := getSockStats("./fake"); err == nil {
```
bowei: uses of the filesystem should be hermetic (create a temporary subdir for this test)

also, probably don't need to check this in the unit test

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	if mcfg.enabledCollectors == "" {
+		glog.Infof("No metrics collectors were enabled.")
+		return nil
+	}
+	enabledCollectors := strings.Split(mcfg.enabledCollectors, ",")
+	nc, pc, err := collector.NewNodeCollector(enabledCollectors, mcfg.procPath)
+	if err != nil {
+		return err
+	}
+	glog.Infof("Enabled metrics collectors:")
```
bowei: glog.Infof("Enabled metrics collectors: %+v", nc.Collectors)

will print out the list of collectors

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	for n, _ := range nc.Collectors {
+		glog.Infof(" - %s", n)
+	}
+
+	registry := prometheus.NewRegistry()
+	err = registry.Register(nc)
+	if err != nil {
+		glog.Infof("Couldn't register collector: %v", err)
```
bowei: Errorf

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil, nil, fmt.Errorf("missing collector: %q", name)
+		} else {
+			if c, err := createFunc(); err == nil {
+				collectors[name] = c
+			} else {
+				return nil, nil, err
+			}
+		}
+	}
+	return &nodeCollector{Collectors: collectors, startTime: time.Now().Unix()},
```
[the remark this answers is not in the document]
lzang: Wouldn't it be too long if using only one line?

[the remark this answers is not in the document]
bowei: break it like this:
```
return &nodeCollector{
  Collectors: ...,
}
```

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+	"io"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type conntrackCollector struct {
+	conntrackEntriesDesc    *prometheus.Desc
```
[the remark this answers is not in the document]
lzang: Just to make them local to this collector, indicating the stats collected by this collector.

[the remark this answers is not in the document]
bowei: if it is const, it's better to leave them global and set them in the file init()

pkg/metrics/collector/sockstat_test.go
lzang: This is not multiple tests. In normal flow,  output is a combination of both ipv4 and ipv6.

bowei: make in a []string then...

pkg/metrics/metrics.go
lzang: nc.Collectors is a map. So your way would print both the name and pointers.

bowei: ok

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+		glog.Infof("No metrics collectors were enabled.")
+		return nil
+	}
+	enabledCollectors := strings.Split(mcfg.enabledCollectors, ",")
+	nc, pc, err := collector.NewNodeCollector(enabledCollectors, mcfg.procPath)
+	if err != nil {
+		return err
+	}
+	glog.Infof("Enabled metrics collectors:")
+	for n, _ := range nc.Collectors {
```
bowei: you don't need the extra underscore

`n := range nc.Collectors {`

pkg/metrics/collector/conntrack.go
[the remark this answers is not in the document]
lzang: It is const but only within this collector, which is local right? The collector variable only inits once.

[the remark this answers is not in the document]
bowei: member variables are used for mutable state. something like a desc that does not change (ever) should be treated as a global const. If this was C++, you would have
```
const conntrackEntriesDesc prometheus::Desc(...);
```
but this is not possible with golang. the next best thing is to create a global var binding and catch violations via code review.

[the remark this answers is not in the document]
lzang: Cannot make them const due to type conflict. The vars are only used by the specific collector, not really global scope. Make them local to the struct has one additional advantage that if the collector is disabled, the object would not be created. So I left it as is.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+	for i, v := range nameParts {
+		switch v {
+		case "insert_failed":
+			indices.insertFailed = i
+		case "drop":
+			indices.drop = i
+		}
+	}
+	if indices.insertFailed == -1 || indices.drop == -1 {
+		return nil, fmt.Errorf("Invalid header %q: don't have target fields", line)
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return indices, nil
+}
+
+// parseConntrackData parses a line of conntrack stat file
+func parseConntrackData(line string, indices *conntrackIndices) (*conntrackStats, error) {
+	stats := &conntrackStats{}
+	valueParts := strings.Split(line, " ")
+	if len(valueParts) != indices.numFields {
+		return nil, fmt.Errorf("Invalid input %q: don't have enough fields", line)
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	if v, err := strconv.ParseUint(valueParts[indices.insertFailed], 16, 32); err == nil {
+		stats.insertFailed = v
+	} else {
+		return nil, fmt.Errorf("Invalid input %q: don't have a valid insertFailed field", line)
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	if v, err := strconv.ParseUint(valueParts[indices.drop], 16, 32); err == nil {
+		stats.drop = v
+	} else {
+		return nil, fmt.Errorf("Invalid input %q: don't have a valid drop field", line)
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// parseConntrackData parses a line of conntrack stat file
+func parseConntrackData(line string, indices *conntrackIndices) (*conntrackStats, error) {
+	stats := &conntrackStats{}
+	valueParts := strings.Split(line, " ")
+	if len(valueParts) != indices.numFields {
+		return nil, fmt.Errorf("Invalid input %q: don't have enough fields", line)
+	}
+
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, fmt.Errorf("Invalid input %q: don't have a valid drop field", line)
+	}
+	return stats, nil
+}
+
+// parseConntrackFile parses the conntrack file contents read in from `r` and returns a merged set of stats.
+func parseConntrackFile(r io.Reader) (*conntrackStats, error) {
+	scanner := bufio.NewScanner(r)
+	if !scanner.Scan() {
+		return nil, fmt.Errorf("Empty file")
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil
+	}
+	ch <- prometheus.MustNewConstMetric(
+		c.conntrackEntriesDesc, prometheus.GaugeValue, float64(value))
+
+	stats, err := getConntrackStats(procFilePath("net/stat/nf_conntrack"))
+	if err != nil {
+		return err
+	}
+	ch <- prometheus.MustNewConstMetric(
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+	defer file.Close()
+	return parseConntrackFile(file)
+}
+
+func (c *conntrackCollector) Update(ch chan<- prometheus.Metric) error {
+	value, err := readUintFromFile(procFilePath("sys/net/netfilter/nf_conntrack_count"))
+	if err != nil {
+		return nil
+	}
+	ch <- prometheus.MustNewConstMetric(
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack.go
```diff
@@ -0,0 +1,161 @@
...
+		c.conntrackEntriesDesc, prometheus.GaugeValue, float64(value))
+
+	stats, err := getConntrackStats(procFilePath("net/stat/nf_conntrack"))
+	if err != nil {
+		return err
+	}
+	ch <- prometheus.MustNewConstMetric(
+		c.conntrackErrorCountDesc, prometheus.CounterValue, float64(stats.insertFailed), "insert_failed")
+	ch <- prometheus.MustNewConstMetric(
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,86 @@
...
+*/
+
+package collector
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestConntrack(t *testing.T) {
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,86 @@
...
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+			_, err := parseConntrackFile(strings.NewReader(tc.input))
+			if err == nil {
+				t.Errorf("parseConntrackFile(%q) = _, nil, want error", tc.input)
+			}
+		})
+	}
+
+	if _, err := getConntrackStats("./fake"); err == nil {
```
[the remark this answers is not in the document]
lzang: Removed the test case.

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"testing"
+)
+
+func TestConntrack(t *testing.T) {
+	input := "entries  searched found new invalid ignore delete delete_list insert insert_failed drop early_drop icmp_error  expect_new expect_create expect_delete search_restart\n" +
+		"000005d2  00000000 00000000 00000000 00000003 00512c38 00000000 00000000 00000000 00000000 00000000 00000000 00000000  00000000 00000000 00000000 00000290\n" +
+		"000005d2  00000000 00000000 00000000 00000002 0047b7ef 00000000 00000000 00000000 00000008 00000005 00000000 00000000  00000000 00000000 00000000 00000255\n" +
+		"000005d2  00000000 00000000 00000000 00000003 0042ab82 00000000 00000000 00000000 00000010 00000000 00000000 00000000  00000000 00000000 00000000 00000245\n" +
+		"000005d2  00000000 00000000 00000000 00000000 00403f1b 00000000 00000000 00000000 00000000 00000020 00000000 00000000  00000000 00000000 00000000 0000026d"
+
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,86 @@
...
+	stats, err := parseConntrackFile(strings.NewReader(input))
+	if err != nil {
+		t.Errorf("parseConntrackFile returns err %v, want nil", err)
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	if stats.insertFailed != 24 {
+		t.Errorf("parseConntrackFile returns insertFailed %v, want 24", stats.insertFailed)
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/conntrack_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	if stats.drop != 37 {
+		t.Errorf("parseConntrackFile returns drop %v, want 37", stats.drop)
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+	"io"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type sockStatCollector struct {
+	socketInuseDesc  *prometheus.Desc
```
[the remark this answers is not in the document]
lzang: Replied above. Leave as is.

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+		socketMemoryDesc: prometheus.NewDesc(
+			"socket_memory",
+			"Amount of memory used by sockets",
+			nil, nil,
+		),
+	}, nil
+}
+
+type socketStats struct {
+	tcpInuse uint64 // num of tcp inuse sockets
```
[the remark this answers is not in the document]
lzang: Updated

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+	udpInuse uint64 // num of udp inuse sockets
+	tcpTw    uint64 // num of tcp tw sockets
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,172 @@
...
+	mem      uint64 // sum of tcp and udp memory
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/sockstat_test.go
lzang: Updated.

pkg/metrics/collector/conntrack_test.go
lzang: Removed the test case.

pkg/metrics/collector/sockstat.go
lzang: Updated.

pkg/metrics/collector/sockstat.go
lzang: Removed.

pkg/metrics/collector/sockstat.go
lzang: Updated.

pkg/metrics/collector/sockstat_test.go
lzang: Change it to []string as the other comment suggested.

pkg/metrics/collector/sockstat.go
lzang: Updated

pkg/metrics/collector/sockstat_test.go
lzang: Updated.

pkg/metrics/collector/sockstat_test.go
lzang: Removed the test case.

pkg/metrics/metrics.go
lzang: Updated.

pkg/metrics/metrics.go
lzang: Updated.

cmd/netd/main.go
```diff
@@ -50,6 +51,11 @@ func main() {
 	glog.Infof("Starting netd")
 	go nc.Run(stopCh, &wg)
 
+	err := metrics.StartCollector()
+	if err != nil {
+		glog.Error("Fail to start metrics collector: %v!", err)
```
bowei: "Could not start metrics collector: %v"

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,126 @@
...
+		if !exist {
+			return nil, nil, fmt.Errorf("missing collector: %q", name)
+		}
+		if c, err := createFunc(); err == nil {
+			collectors[name] = c
+		} else {
+			return nil, nil, err
+		}
+	}
+	return &nodeCollector{
```
bowei: one line...

pkg/metrics/collector/conntrack.go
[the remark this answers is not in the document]
bowei: I know it cannot be const, that was just an example from a different language -- I'm saying the pattern in go is to make these a `var` at the global level.

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,185 @@
...
+	"io"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type sockStatCollector struct {
+	socketInUseDesc    *prometheus.Desc
```
bowei: same as above

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,185 @@
...
+	file, err := os.Open(fileName)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	return parseSockStats(file)
+}
+
+func (s *socketStats) merge(o *socketStats) {
```
bowei: move this above getSocketStats()

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,185 @@
...
+	out     interface{}
+	extract func(string, string, interface{}) error
+}
+
+func extractSockStatFields(fields []string, entries map[string]*sockStatEntry) error {
+	for i := 0; i < len(fields) && i+1 < len(fields); i += 2 {
+		k, v := fields[i], fields[i+1]
+		if e, ok := entries[k]; !ok {
+			continue
+		} else if err := e.extract(k, v, e.out); err != nil {
```
bowei: you don't need else if if you have a `continue` above

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+	if lineCount == 0 {
+		return nil, fmt.Errorf("empty file")
+	}
+	return ret, nil
+}
+
+func (c *sockStatCollector) Update(ch chan<- prometheus.Metric) error {
+	ss, err := getSockStats(procFilePath("net/sockstat"))
```
bowei: Why not pass location of sockstat and sockstat6 as flags with default value `/proc/net/{sockstat,sockstat6}?`
What is the purpose of procFilePath?

pkg/metrics/collector/sockstat.go
```diff
@@ -0,0 +1,185 @@
...
+			nil, nil,
+		),
+	}, nil
+}
+
+type socketStats struct {
+	tcpInUse        uint64 // num of tcp in use sockets
+	udpInUse        uint64 // num of udp in use sockets
+	tcpTimeWait     uint64 // num of tcp timewait sockets
+	memUsedInBlocks uint64 // sum of tcp and udp memory used in blocks
```
bowei: "blocks" probably should be "page size"

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+		glog.Errorf("Couldn't register collector: %v", err)
+		return err
+	}
+
+	for _, c := range pc {
+		registry.MustRegister(c)
+	}
+
+	gatherers := prometheus.Gatherers{
+		//prometheus.DefaultGatherer,
```
bowei: remove commented out code

cmd/netd/main.go
lzang: Updated.

pkg/metrics/collector/collector.go
lzang: Updated. Cannot put the whole return in one line as it looks too long.

pkg/metrics/collector/sockstat.go
lzang: Updated

pkg/metrics/collector/sockstat.go
lzang: Used else because otherwise cannot use the variable e. Changed the code to increase the scope of variable e as the following
e, ok := entries[k]
if !ok {
	continue
} 
if err := e.extract(k, v, e.out); err != nil {

pkg/metrics/collector/sockstat.go
lzang: Changed it to memUsedInPages

pkg/metrics/metrics.go
lzang: Left it so that it was easy to be enabled for debugging. Removed it in the updated patch for cleanness.

pkg/metrics/collector/sockstat.go
lzang: Accidentally hit resolve for the comment above, so will reply here. Understand your points in general. I would think one advantage to make them local is that when they are disabled, no global object would be created. If you still want to make them global, let me know, I can update it.

pkg/metrics/collector/sockstat.go
lzang: Because there is no point to have individual flags when many collectors share the same proc directory. The reason to have the procFilePath() is that it is easy to change proc path for all. This is used when the program is running in a container, so that the host proc can be mapped to a different directory inside the container for the program to run correctly.

pkg/metrics/collector/sockstat.go
bowei: global object should not be a problem...

lzang: OK, updated to using global vars.

cmd/netd/main.go
```diff
@@ -50,6 +51,11 @@ func main() {
 	glog.Infof("Starting netd")
 	go nc.Run(stopCh, &wg)
 
+	err := metrics.StartCollector()
+	if err != nil {
+		glog.Error("Could not start metrics collector: %v!", err)
```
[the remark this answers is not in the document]
lzang: Updated.

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,123 @@
...
+var (
+	scrapeFailureCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "scrape_collector_fail",
+		Help: "scrape collector fail count.",
+	}, []string{"collector"})
+
+	scrapeDurationHist = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "scrape_collector_durations_histogram_seconds",
+		Help:    "scrape duration distributions.",
+		Buckets: prometheus.ExponentialBuckets(0.00005, 4, 8),
```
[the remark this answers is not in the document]
lzang: Updated

pkg/metrics/collector/collector.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		glog.Errorf("Collector %q failed (took %v) %v", name, duration, err)
+		scrapeFailureCount.WithLabelValues(name).Inc()
+	}
+}
+
+// Collector is the interface a collector has to implement.
+type Collector interface {
+	// Get new metrics and expose them via prometheus registry.
+	Update(ch chan<- prometheus.Metric) error
```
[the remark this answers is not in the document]
lzang: Updated.

PR #53, opened 2018-10-29
Enable IPv6 forwarding when Calico CNI is used.

jingax10:
It implicitly requires the container to have capability to run node-level sysctl cmds.


scripts/install-cni.sh
```diff
@@ -52,6 +52,10 @@ if [ "$ENABLE_PRIVATE_IPV6_ACCESS" == "true" ]; then
     cni_spec=$(echo ${cni_spec:-} | sed -e \
       "s#@ipv6SubnetOptional#, [{\"subnet\": \"${node_ipv6_addr:-}/112\"}]#g;
        s#@ipv6RouteOptional#, {\"dst\": \"::/0\"}#g")
+    if [ "${ENABLE_CALICO_NETWORK_POLICY}" == "true" ]; then
+      echo "Enabling IPv6 forwarding..."
+      sysctl -w net.ipv6.conf.all.forwarding=1
```
cadmuxe: Agree that we shouldn't do it in here. But where is the change in netd? do we have a bug to tracking this?

jingax10: We have an internal bug (already cced you) to track the issue.

PR #58, opened 2019-01-23
Update netd metrics default config and yaml

lzang:


pkg/metrics/metrics.go
```diff
@@ -86,6 +86,7 @@ func StartCollector() error {
 		if err != nil {
 			glog.Errorf("Couldn't start http server- %v", err)
 		}
+		return
```
cadmuxe: is this necessary?

Issue #33, opened 2018-09-12
Convert scripts/install-cni.sh to a golang


jingax10:
It is hard to unit test CNI generator by scripts/install-cni.sh. We want to convert to golang to make it unit testable.

rlenglet:
@varunmar Can you please reassign this issue to @bcheung? He who will start working on it soon.

varunmar:
Hmm. I think I can't assign it to him directly. I wonder if it restricts it to the GoogleCloudPlatform org or something

PR #59, opened 2019-01-31
Add required IPv6 firewall rules when IPv6 support is enabled

varunmar:


varunmar:
I think it only needs to be done once (per node, on boot) - Are you recommending we run this inside the reconcile loop?
That's not what the IPv4 versions of this do - they're all in a startup script. Eg. https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/configure-helper.sh#L69

cadmuxe:
Ok, If it's one time setup, we can keep it in here. And I am planning to change the bash to go code for better testing and management. 
Does this in your design doc?

varunmar:
> install-cni.sh is triggered by init container which uses netd image. So if you make any changes to this script, the netd image needs to be rebuilt to include the updated script.
Yes indeed, we will need to spin up a new version so Directpath enabled clusters can use IPv6.

> On the other hand, please put comments on why you need to install these allow rules and why the rules look different from the IPv4 rules.
They're actually the same as the IPv4 rules, just translated to IPv6 packets. I added a comment saying so.

> Ok, If it's one time setup, we can keep it in here. And I am planning to change the bash to go code for better testing and management.
> Does this in your design doc?
SG - yes, a move to golang would be great. I think we already have an open issue for that, feel free to take it if you want - https://github.com/GoogleCloudPlatform/netd/issues/33

PR #61, opened 2019-05-14
Allow two authentication strategies for requests to Kubernetes APIServer

haiyanmeng:
Currently, the requests from `install_cni.sh` to Kubernetes APIServer are all authenticated with a [service account token](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens).
This PR adds support for [client certificate authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#x509-client-certs). With client certification authentication, `install-cni.sh` will use kubelet's client certificate for authentication.

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -34,8 +34,21 @@ fi
 
 # Fill CNI spec template.
 token=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+kubelet_client_cert=/var/lib/kubelet/pki/kubelet-client-current.pem
+kubelet_client_key=/var/lib/kubelet/pki/kubelet-client-current.pem
+ca_cert=/etc/srv/kubernetes/pki/ca-certificates.crt
 node_url="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
-ipv4_subnet=$(curl -k -s -H "Authorization: Bearer $token" $node_url | jq '.spec.podCIDR')
+ipv4_subnet=""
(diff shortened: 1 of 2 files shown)


cadmuxe:
Hi Haiyan, Thank you for helping us improve the security. Maybe also clean the yaml in the same PR? 
Thanks

scripts/install-cni.sh
```diff
@@ -35,7 +35,20 @@ fi
...
 token=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
 node_url="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
-ipv4_subnet=$(curl -k -s -H "Authorization: Bearer $token" $node_url | jq '.spec.podCIDR')
+ipv4_subnet=""
+if [ -n "${token}" ]; then
+  # use service account token authentication strategy
+  # a service account token should be mounted into the container for this option
+  ipv4_subnet=$(curl -k -s -H "Authorization: Bearer $token" $node_url | jq '.spec.podCIDR')
+else
+  # use client certification authentication strategy
```
tallclair: Should we prefer this strategy, or maybe make the choice explicit with an environment variable? E.g.
```
if [ -n "${CLIENT_KEY_PATH}" ] && [ -n "${CLIENT_CERT_PATH}" ]; then
   ...
```

haiyanmeng:
> Maybe also clean the yaml in the same PR?

Done

scripts/install-cni.sh
haiyanmeng: Good point: the client cert authentication should be preferred. Fixed it!

We may still want to set `AutomountServiceAccountToken` to `false` in a future PR after we verify that the client cert authentication works.

PR #79, opened 2020-02-04
update init script 

cadmuxe:


scripts/install-cni.sh
```diff
@@ -17,16 +17,16 @@
 # overide calico network policy config if its cni is not installed as expected
 echo "Calico network policy config: " $ENABLE_CALICO_NETWORK_POLICY
 if [[ ${ENABLE_CALICO_NETWORK_POLICY} == "true" ]]; then
-  cni_file=`ls /host/etc/cni/net.d/*calico* 2> /dev/null`
+  cni_file=`ls /host/etc/cni/net.d/*calico*.conflist 2> /dev/null`
```
cadmuxe: /host/etc/cni/net.d/*calico* 
will also select the calico-kubeconfig under cni/netd.d folder

scripts/install-cni.sh
```diff
@@ -17,16 +17,16 @@
...
   WAITED=0
-  while [[ ${cni_file} == ""  && ${WAITED} -lt 120 ]]
```
cadmuxe: This will generate an error msg: sh: : unknown operand which will break the script.

PR #81, opened 2020-02-20
Adds an env var to enable chaining Cilium plugin.

Weil0ng:
Adds a handle for new env `ENABLE_CILIUM_PLUGIN`, which chains cilium cni at `@cniCiliumPlugin` annotation.

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -64,6 +64,13 @@ else
 	cni_spec=$(echo ${cni_spec:-} | sed -e "s#@cniBandwidthPlugin##g")
 fi
 
+if [ "${ENABLE_CILIUM_PLUGIN}" == "true" ]
+then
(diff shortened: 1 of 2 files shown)


scripts/install-cni.sh
```diff
@@ -64,6 +64,13 @@ else
 	cni_spec=$(echo ${cni_spec:-} | sed -e "s#@cniBandwidthPlugin##g")
 fi
 
+if [ "${ENABLE_CILIUM_PLUGIN}" == "true" ]
+then
+        cni_spec=$(echo ${cni_spec:-} | sed -e "s#@cniCiliumPlugin#,{\"type\": \"cilium-cni\"}#g")
```
cadmuxe: cloud you also update the netd.yaml, we are using it as a reference. 

https://github.com/GoogleCloudPlatform/netd/blob/master/netd.yaml#L71

Weil0ng: Included another commit for this :)

Weil0ng:
PTAL :) @cadmuxe

PR #83, opened 2020-06-15
Write generated cni config file atomically

bcheung:
This just ensures that the resulting CNI config file is generated then renamed atomically.

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -183,8 +183,12 @@ else
   echo "Creating CNI spec..."
 fi
 
-cat > "${output_file}.tmp" <<EOF
-${cni_spec:-}
-EOF
-
-mv "${output_file}.tmp" ${output_file}
+# Atomically write CNI spec
+temp_file=$(mktemp ${output_file}.tmp.XXXXXX)
+if [ $? -ne 0 ]; then
+  echo "Failed to create temp file, Exiting (1)..."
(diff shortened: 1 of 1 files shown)


scripts/install-cni.sh
```diff
@@ -183,6 +183,8 @@ else
   echo "Creating CNI spec..."
 fi
 
-cat >${output_file} <<EOF
+cat > "${output_file}.tmp" <<EOF
```
rlenglet, line 186: You may want to use `mktemp` to create the tempfile, which should be available on Alpine.
Also, it would be good form to delete the tempfile unconditionally with a `trap`, e.g.
```
trap "rm -f $tempfile" EXIT
```

rlenglet: LGTM, thanks!

bcheung:
Could you approve the changes with the necessary flags please? @lzang 
@rlenglet told me about the issues that could arise with using echo instead of cat, so I reverted it back.
(I should also probably leave the existing code the same unless there is actually an issue because it may have been intentional to use cat over echo originally 😄)

Here's some more context (near the end of the accepted answer): [Difference between cat and echo](https://superuser.com/questions/1434549/whats-the-different-between-cat-some-file-eof-some-stuff-eof-and-echo)

bcheung:
@lzang I think I need a lgtm label?

PR #85, opened 2020-07-27
Migrate to go modules

bcheung:
Migrate to Go modules (Go v1.14) and removes usage of https://github.com/golang/dep locally and in Travis CI. This PR still uses the vendor directory.

Running `go mod vendor` would download the dependencies into the module cache and copy them into the vendor folder which will be used to build and develop locally. ([Go v1.14](https://golang.org/doc/go1.14#go-command) supports vendoring if a vendor directory is present)

Ensuring that the vendor directory is up to date and valid (see [Golang Vendoring](https://golang.org/ref/mod#vendoring)):
> When the `go` command reads vendor/modules.txt, it checks that the module versions are consistent with go.mod. If go.mod has changed since vendor/modules.txt was generated, the go command will report an error. go mod vendor should be run again to update the vendor directory.

See [README.vendor](https://golang.org/src/README.vendor) for instructions to add dependencies to the project.

Notes about the the current build process and why vendoring is still used:
The current build process uses a docker container to build the netd binary. This keeps the cache and dependencies self-contained in the project directory and does not affect the user's development environment. This build process and removal of the vendor directory has some advantages and drawbacks: 

Pros:
- A self-contained allows us to build the binaries without making too many assumptions about the developer's development environment.
- Go is moving in the direction of supporting Go modules. (moving forward, we may be able to simplify the build process if Go introduces more support)

Cons:
- Not using a vendor directory introduces redundancy in downloading the module cache twice. The build process downloads the module cache to the .go scratch directory (for the sole purpose of building the binary), but to develop Go locally, the developer would also need to download the dependencies to their module cache in `$GOPATH/go/pkg/mod`.


anfernee:
@bcheung can you explain why `go mod vendor` is needed

bcheung:
> @bcheung can you explain why `go mod vendor` is needed

I ended up moving away from vendoring, but it is still supported. (added to the PR description)

anfernee:
@lzang can you take a look  at this?

lzang:
Are you removing the vendor files so that builder needs to generator the vendor files on the fly? But why? Even with go modules, we can still have a vendor directory. I don't see the advantage of removing them but it has the cons you already mentioned.

bcheung:
> Are you removing the vendor files so that builder needs to generate the vendor files on the fly? But why? Even with go modules, we can still have a vendor directory. I don't see the advantage of removing them but it has the cons you already mentioned.

@lzang Good point it may be a bit early to adopt Go modules without vendoring at the moment. This was more of an open proposal to be considered. I just saw many repos and online sources trying to move away from vendoring, but might be a good idea to just stick with vendoring for now.

lzang:
Thanks Brian!

PR #86, opened 2020-07-28
Add golanci-lint and fix errors

bcheung:
Adds golanci-lint make target and adds the command to Travis CI.
Fixes lint errors and a bug with the `reconcile-interval-seconds` flag in the netd binary.


Makefile
```diff
@@ -195,6 +197,15 @@ shell: init
 	@echo "launching a shell in the containerized build environment"
 	@$(DOCKER_RUN) /bin/sh $(CMD)
 
+.PHONY: lint-go
+lint-go:
+	@docker run												\
+		--rm 												\
+		-v "$$(pwd):/app"									\
+		-w /app												\
+		golangci/golangci-lint:$(GOLANGCI_LINT_VERSION)		\
```
lzang, line 206: Thanks for the work! What is the difference of golangci-lint vs. the ordinary lint? I also see you have a .golangci.yml, which includes what would be checked or not checked - how is that decided?

PR #88, opened 2020-07-30
Improve build pipeline

bcheung:
Cleans up Makefile and runs the build container as non-root user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,10 @@
 # OSX trash
 .DS_Store
 
+# Intellij files
+*.iml
+.idea/
+
 # Vscode files
 .vscode
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: required
 services:
   - docker
 language: go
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: required
```
anfernee, line 1: this is great!!

Makefile
```diff
@@ -42,22 +42,35 @@ ALL_ARCH := amd64 arm arm64 ppc64le
...
 
 IMAGE := $(REGISTRY)/$(BIN)-$(ARCH)
 
 BUILD_IMAGE ?= golang:1.14-alpine
 
+# Docker run command prefix
+DOCKER_RUN =  docker run                                                            \
+				--rm                                                                \
+				-u $$(id -u):$$(id -g)                                              \
+				-v "$$(pwd)/.go/.cache:/.cache"                                     \
```
anfernee, line 65: `$(CURDIR)`` as opposed to shelling out?

anfernee:
Look great. Can you also add `.PHONY` for all the non-file targets?

anfernee:
@lzang can you take a look?

.gitignore
```diff
@@ -4,6 +4,10 @@
 # OSX trash
 .DS_Store
 
+# Intellij files
```
lzang, line 7: Why do you have these files?

bcheung: These are just files that Intellij Goland IDE generates. (like the Vscode files)

PR #89, opened 2020-08-05
Remove unnecessary calico template variables

bcheung:


scripts/install-cni.sh
```diff
@@ -37,19 +37,20 @@ set -u -e
...
 if [ "${ENABLE_CALICO_NETWORK_POLICY}" == "true" ]; then
   echo "Calico Network Policy is enabled"
-  if [ -z "${CALICO_CNI_SPEC_TEMPLATE_FILE:-}" ]; then
-    echo "No Calico CNI spec template is specified. Exiting (0)..."
+  # Backwards compatibility support
+  if [ -n "${CALICO_CNI_SPEC_TEMPLATE_FILE}" ] && [ -n "${CALICO_CNI_SPEC_TEMPLATE}" ]; then
+    # Use environment overrides
+    calico_cni_file="${CALICO_CNI_SPEC_TEMPLATE_FILE}"
+    cni_spec="${CALICO_CNI_SPEC_TEMPLATE}"
+  elif [ "${OVERWRITE_CALICO_CNI_FILE:-"false"}" == "false" ]; then
```
anfernee, line 45: This skip all the iptables rules, is that the intention?

bcheung: Yes, in the original logic, if `CALICO_CNI_SPEC_TEMPLATE_FILE` and `CALICO_CNI_SPEC_TEMPLATE` were not set, then the script exits.

bcheung:
These changes did not take into account the case where Calico CNI is not correctly installed (likely due to the behavior discussed in #91).
This is why the Calico CNI template variables are needed (`CALICO_CNI_SPEC_TEMPLATE_FILE` and `CALICO_CNI_SPEC_TEMPLATE`) so that the script can use the normal CNI template if it overrides `ENABLE_CALICO_NETWORK_POLICY` and sets it to false.

PR #99, opened 2020-09-01
Do not wait for calico conf file if ipv6 support is requested.

### LINKED ISSUE #98: When netd is being used to install the cni config template for direct IPv6 support, do not wait for the config file to be written first
When ENABLE_PRIVATE_IPV6_ACCESS is set along with ENABLE_CALICO_NETWORK_POLICY, the netd install-cni.sh script is used to initially generate the template that the calico install-cni container will use to generate the complete cni config.

In such cases (ie, both ENABLE_PRIVATE_IPV6_ACCESS and ENABLE_CALICO_NETWORK_POLICY are true, we should not wait for the config file to be written.

varunmar:
Fixes #98

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -14,9 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# overide calico network policy config if its cni is not installed as expected
+# Overide calico network policy config if its cni is not installed as expected.
+# This can happen if the calico daemonset is removed but the master addon still exists.
+#
+# If this script is being run in order to generate the Calico config file, then skip this
+# check.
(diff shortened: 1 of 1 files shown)


scripts/install-cni.sh
```diff
@@ -16,7 +16,7 @@
 
 # overide calico network policy config if its cni is not installed as expected
 echo "Calico network policy config: " $ENABLE_CALICO_NETWORK_POLICY
-if [[ ${ENABLE_CALICO_NETWORK_POLICY} == "true" ]]; then
+if [[ ${ENABLE_CALICO_NETWORK_POLICY} == "true" ]] && [ "$ENABLE_PRIVATE_IPV6_ACCESS" != "true" ]; then
```
varunmar, line 19: WRITE_CALICO_TEMPLATE instead

PR #106, opened 2021-04-06
Add masks to CONNMARK save/restore operations.

### LINKED ISSUE #105: netd clobbers the whole FW mark when restoring the CONNMARK
While testing Calico with intra-node visibility enabled, we spotted a subtle incompatibility:

* Calico uses mark bits to communicate between the various iptables chains.  For example, if "pre-DNAT" policy accepts a packet then a flag bit is set in the mangle-PREROUTING chain to tell the downstream "normal" policy that the packet was pre-approved by the earlier policy.

* netd stores and retrieves mark bits in the connection tracking mark.  While it only seems to use one or two bits of the connmark, it uses a `--restore -mark` operation with no mask to copy the connmark to the fw mark.  Since Calico's mark bits are per-packet and we don't store them in the connmark, this restore operation can clobber Calico's flag bits.

So, if the user uses certain combinations of Calico policy and intra-node visibility, they can get unexpected results.  Calico's eBPF dataplane also uses mark bits per-packet and these can also be clobbered.

I suggest using a mask on the restore operation so that Calico and netd can each manage their own ranges of bits.

fasaxc:
Avoids clashes with other uses of the mark bits.

Fixes #105.

### DIFF
--- a/pkg/config/config_policy_routing.go
+++ b/pkg/config/config_policy_routing.go
@@ -89,6 +89,7 @@ func init() {
 	localLinkIndex, localNetdev, _ = f(net.IPv4(127, 0, 0, 1))
 
 	sysctlReversePathFilter := fmt.Sprintf("net.ipv4.conf.%s.rp_filter", defaultNetdev)
+	hairpinMaskStr := fmt.Sprintf("0x%x", hairpinMask)
 	PolicyRoutingConfigSet.Configs = []Config{
 		SysctlConfig{
 			Key:          sysctlReversePathFilter,
@@ -110,7 +111,10 @@ func init() {
 				IPT:            ipt,
 			},
 			[]IPTablesRuleSpec{
-				[]string{"-j", "CONNMARK", "--restore-mark", "-m", "comment", "--comment", policyRoutingGcpPreRoutingComment},
+				[]string{
+					"-j", "CONNMARK", "--restore-mark", "--nfmask", hairpinMaskStr, "--ctmask", hairpinMaskStr,
+					"-m", "comment", "--comment", policyRoutingGcpPreRoutingComment,
+				},
 			},
 			ipt,
 		},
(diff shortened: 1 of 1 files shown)


MrHohn (commented):
Thanks for the PR @fasaxc ! I don't see any problem, but would like to test it out end-to-end before merging. Do you have a quick test case in mind? e.g. what kind of networkpolicy config was broken previously?

fasaxc:
I was testing Calico's eBPF dataplane option by spinning up a cluster with intra-node visibility (to get GKE CNI) and then applying Calico manually on top of that.  However, when I worked out what was wrong, I realised it would affect other scenarios too, including Calico's host endpoint policy, which allows you to apply Calico policy rules to the host NIC.  In particular, I think it affected Calico's pre-DNAT policy, which is useful for securing NodePorts, and our do-not-track policy.

Those policy types set flag bits to mark packets as "pre-approved" in the  `mangle/PREROUTING` and `raw/PREROUTING` chains, which then cause policy to be skipped in the `filter/FORWARD` chain.  If the mark bits get overwritten then the pre-DNAT policy might allow a packet only to have it dropped by the "normal" policy later.  To see the problem, I think you'd need to create:

* A Calico GlobalNetworkPolicy with `preDNAT: true` that allowed allows certain flows (e.g. allow ICMP)
* A Calico GlobalNetworkPolicy with `applyOnForward: true` that drops the same flows (e.g. drop ICMP) but allows everything else (to avoid cutting off the host completely)
* A host endpoint with the two policies above applied to it.

With those policies in place, the preDNAT policy should "win" and allow ICMP to the host but I think we'll apply both policies and do the wrong thing.

fasaxc:
@googlebot I fixed it.

fasaxc:
Looks like CI failed with a docker pull rate limit?

fasaxc:
@MrHohn Please can you rerun the CI?  Looks like it failed with a dockerhub quota issue.

MrHohn:
@fasaxc Yes, I have requested it to re-run in the right way (previously requested on github and it somehow didn't go through). On the other hand, I have tested this patch locally and it seems to be working properly. I will merge this patch and include it in the next release.

MrHohn (approved):
/lgtm
/approve

fasaxc:
Thanks @MrHohn; can you give me any insight into how netd releases flow through into GKE releases?  When might this be available to try in GKE?

MrHohn:
@fasaxc We will need to release a new version of netd that contains the fix, then update the internal mapping so GKE can pick it up. I would expect <1 month to have something triable in GKE (e.g. rapid channel).

fasaxc:
Great, thanks, have set myself a reminder to check back in a month.

PR #107, opened 2021-04-12
netd changes to support dual stack (ipv4_ipv6)

basantsa1989:
When IPV6 is enabled, we'll populate the ipv6Subnet from the value set in spec.podCIDRs[1]

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -128,7 +128,7 @@ if [ "$ENABLE_PRIVATE_IPV6_ACCESS" == "true" ] || [ "$ENABLE_IPV6" == "true" ]; 
     echo "Found nic0 IPv6 address ${node_ipv6_addr:-}. Filling IPv6 subnet and route..."
 
     if [ "$ENABLE_IPV6" == "true" ]; then
-      ipv6_subnet=$(echo $response | jq '.spec.podCIDRs[1]')
+      ipv6_subnet=${node_ipv6_addr%:*:*}:ffff:/112
       echo "IPv6 subnet: ${ipv6_subnet:-}"
       cni_spec=$(echo ${cni_spec:-} | sed -e \
         "s#@ipv6SubnetOptional#, [{\"subnet\": ${ipv6_subnet:-}}]#g;
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -127,16 +127,9 @@ if [ "$ENABLE_PRIVATE_IPV6_ACCESS" == "true" ] || [ "$ENABLE_IPV6" == "true" ]; 
   if [ -n "${node_ipv6_addr:-}" ] && [ "${node_ipv6_addr}" != "null" ]; then
     echo "Found nic0 IPv6 address ${node_ipv6_addr:-}. Filling IPv6 subnet and route..."
 
-    if [ "$ENABLE_IPV6" == "true" ]; then
-      ipv6_subnet=${node_ipv6_addr%:*:*}:ffff:/112
-      cni_spec=$(echo ${cni_spec:-} | sed -e \
-        "s#@ipv6SubnetOptional#, [{\"subnet\": ${ipv6_subnet:-}}]#g;
-        s#@ipv6RouteOptional#, {\"dst\": \"::/0\"}#g")
-    else
-      cni_spec=$(echo ${cni_spec:-} | sed -e \
-        "s#@ipv6SubnetOptional#, [{\"subnet\": \"${node_ipv6_addr:-}/112\"}]#g;
-        s#@ipv6RouteOptional#, {\"dst\": \"::/0\"}#g")
-    fi
+    cni_spec=$(echo ${cni_spec:-} | sed -e \
+      "s#@ipv6SubnetOptional#, [{\"subnet\": \"${node_ipv6_addr:-}/112\"}]#g;
+      s#@ipv6RouteOptional#, {\"dst\": \"::/0\"}#g")
 
     # Ensure the IPv6 firewall rules are as expected.
     # These rules mirror the IPv4 rules installed by kubernetes/cluster/gce/gci/configure-helper.sh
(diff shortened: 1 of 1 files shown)


scripts/install-cni.sh
```diff
@@ -121,15 +121,24 @@ if [ "$ENABLE_MASQUERADE" == "true" ]; then
...
 
   if [ -n "${node_ipv6_addr:-}" ] && [ "${node_ipv6_addr}" != "null" ]; then
     echo "Found nic0 IPv6 address ${node_ipv6_addr:-}. Filling IPv6 subnet and route..."
-    cni_spec=$(echo ${cni_spec:-} | sed -e \
-      "s#@ipv6SubnetOptional#, [{\"subnet\": \"${node_ipv6_addr:-}/112\"}]#g;
-       s#@ipv6RouteOptional#, {\"dst\": \"::/0\"}#g")
-    
+
+    if [ "$ENABLE_IPV6" == "true" ]; then
+      ipv6_subnet=$(echo $response | jq '.spec.podCIDRs[1]')
```
MrHohn, line 131: Which component is responsible to fill in `.spec.podCIDRs[1]`? Should the script exit and retry if this is not there yet?

scripts/install-cni.sh
```diff
@@ -121,15 +121,24 @@ if [ "$ENABLE_MASQUERADE" == "true" ]; then
   fi
 fi
 
-if [ "$ENABLE_PRIVATE_IPV6_ACCESS" == "true" ]; then
+if [ "$ENABLE_PRIVATE_IPV6_ACCESS" == "true" ] || [ "$ENABLE_IPV6" == "true" ]; then
   node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq -r '.ipv6s[0]' ) ||:
 
   if [ -n "${node_ipv6_addr:-}" ] && [ "${node_ipv6_addr}" != "null" ]; then
```
MrHohn, line 127: Just double checking - does the same `node_ipv6_addr` checks make sense for dualstack? (Asking as I don't see `node_ipv6_addr` being used anymore.)

scripts/install-cni.sh
basantsa1989: I am working on the k8s changes required to populate .spec.podCIDRs[1].

The changes will be in:
https://github.com/kubernetes/legacy-cloud-providers/blob/master/gce/gce_instances.go#L495
and 
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/nodeipam/ipam/cloud_cidr_allocator.go#L284

With those changes, we should always have a value in .spec.podCIDRs[1] when IPV6/DualStack is enabled.

scripts/install-cni.sh
basantsa1989: When enable_ipv6 is true, we need to have a valid node_ipv6_addr. So the node_ipv6_addr checks are still valid.
Just the way we get ipv6Subnet will be different.

scripts/install-cni.sh
sdmodi: Basant I was testing this and it seems like there is a race in the GCE metadata server and populating the interface IP. Basically what we end up writing in the podCIDRs is sometimes wrong and stale. For now we can continue using the interface's IP address to write the CNI config just like you had suggested. This would look like this:

s#@ipv6SubnetOptional#, [{\"subnet\": \"${node_ipv6_addr%:*:*}:ffff:0/112\"}]#g;

If it is safe we should also do the same thing for directpath. We cannot use the first /112 because then we will conflict with the node IP. When dual stack clusters are enabled, we want to use the last /112 for pods.

basantsa1989: @sdmodi I'll make this change and test. Thanks

basantsa1989: Done. Made the change and tested only for IPV6 Dual Stack for now. 
@varunmar @MrHohn do you see any issues with changing directpath to use last /112 block instead of first /112 block?

MrHohn: Not an expert on the directpath usage with pod - will defer to @varunmar to respond :)

scripts/install-cni.sh
```diff
@@ -121,15 +121,24 @@ if [ "$ENABLE_MASQUERADE" == "true" ]; then
...
     echo "Found nic0 IPv6 address ${node_ipv6_addr:-}. Filling IPv6 subnet and route..."
-    cni_spec=$(echo ${cni_spec:-} | sed -e \
-      "s#@ipv6SubnetOptional#, [{\"subnet\": \"${node_ipv6_addr:-}/112\"}]#g;
-       s#@ipv6RouteOptional#, {\"dst\": \"::/0\"}#g")
-    
+
+    if [ "$ENABLE_IPV6" == "true" ]; then
+      ipv6_subnet=${node_ipv6_addr%:*:*}:ffff:/112
+      echo "IPv6 subnet: ${ipv6_subnet:-}"
```
sdmodi, line 132: This looks like a debug log. We should remove it.

sdmodi (approved):

scripts/install-cni.sh
basantsa1989: Done.

scripts/install-cni.sh
varunmar: I'm not sure  I understand the question - the first/last 112 block of the node allocation? How big is the range handed out to the node - it used to only be /112, is it bigger now?

basantsa1989: Node is assigned a /96 block. Node uses the first address in this block for its interface. So we need to pick a /112 block from it. I am picking the last block to avoid any ip conflicts. Want to check if this applies for directpath too.

basantsa1989: Made it consistent with directpath. Since the node address is :0 and the first address allocated to pods starts from :1, we should not see any conflicts.

varunmar:
/approve cancel
/lgtm

Please squash the commits before merging.

PR #109, opened 2021-06-11
Add a container image for installing random dns delay.

MrHohn:
This is to mitigate a kernel race condition bug where conntrack may
be droping packets silently if two DNS queries (one for IPv4, one
for IPv6) are sent in parallel.

This is largely based of the workaround suggested on:
https://blog.quentin-machu.fr/2018/06/24/5-15s-dns-lookups-on-kubernetes/

/assign @prameshj @sypakine

### DIFF
--- a/scripts/dns-delay/install-dns-delay.sh
+++ b/scripts/dns-delay/install-dns-delay.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+echo "Force the kernel to re-create the dummy mq scheduler on the default interface."
+sysctl -w net.core.default_qdisc=fq_codel
+tc qdisc del dev "$(route | grep '^default' | grep -o '[^ ]*$')" root 2>/dev/null || true
+tc qdisc add dev "$(route | grep '^default' | grep -o '[^ ]*$')" root handle 0: mq || true
+echo "Waiting for eth0 interface to come up."
+while ! ip link | grep "eth0:" > /dev/null; do sleep 1; done
+tc qdisc del dev eth0 root 2>/dev/null || true
+tc qdisc add dev eth0 root handle 1: prio bands 2 priomap 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1
+tc qdisc add dev eth0 parent 1:2 handle 12: fq_codel
+echo "Injecting a small delay using netem."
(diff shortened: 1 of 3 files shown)


sypakine (approved):

scripts/dns-delay/install-dns-delay.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
prameshj, line 1: Do you need the sysctl command too "sysctl -w net.core.default_qdisc=fq_codel" for GCE?

scripts/dns-delay/install-dns-delay.sh
```diff
@@ -0,0 +1,9 @@
...
+tc qdisc del dev "$(route | grep '^default' | grep -o '[^ ]*$')" root 2>/dev/null || true
+tc qdisc add dev "$(route | grep '^default' | grep -o '[^ ]*$')" root handle 0: mq || true
+tc qdisc del dev eth0 root 2>/dev/null || true
```
prameshj, line 4: Do we expect eth0 to be up by the time this script runs? If not, we might need "while ! ip link | grep "eth0:" > /dev/null; do sleep 1; done"

I am not an expert on this.. just comparing the workaround in the blog and here :)

prameshj (commented):
Looks good to me, except 2 small comments.

scripts/dns-delay/install-dns-delay.sh
MrHohn: Good catch, I was omitting it initially just to make sure the rest works. I will add it.

scripts/dns-delay/install-dns-delay.sh
MrHohn: Most likely eth0 will be up by the time netd starts - but no harm to add that :) Will do.

scripts/dns-delay/install-dns-delay.sh
MrHohn: Added now.

scripts/dns-delay/install-dns-delay.sh
MrHohn: Suggestion applied.

MrHohn:
Sorry for the delay - addressed the comments and added a few logs in the script for better debuggability.

MrHohn:
Per the LGTM, merging this to move things forward :)

PR #110, opened 2021-07-13
Kernel metrics

michaelasp:
Add kernel metrics to netd, scraping from nstat and netstat. Still a WIP since some metrics need to be agreed upon and hammered out. 

Adds 5 new metrics inside of the kernel_metrics collector which will be used to monitor node traffic patterns.

### DIFF
--- a/netd.yaml
+++ b/netd.yaml
@@ -108,7 +108,7 @@ metadata:
   labels:
     addonmanager.kubernetes.io/mode: EnsureExists
 data:
-  metrics_collectors: "conntrack,socket"
+  metrics_collectors: "conntrack,socket,kernel_metrics"
   metrics_address: "localhost:10231"
 ---
 
--- a/pkg/metrics/collector/kernel_metrics.go
+++ b/pkg/metrics/collector/kernel_metrics.go
@@ -1,0 +1,155 @@
+/*
+Copyright 2021 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package collector
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	netstatLabel = "TcpExt:" // Label we use to parse netstat tcp metrics
+	snmpLabel    = "Tcp:"    // Label we use to parse snmp tcp metrics
+)
+
+// Keys for metrics after parsing
+const (
+	// Got from Snmp counters
+	tcpSegsIn      = "InSegs"      // Total tcp segments in
+	tcpSegsOut     = "OutSegs"     // Total tcp segments out
+	tcpSegsRetrans = "RetransSegs" // Total tcp segments retransmitted
+	// Got from netstat
+	tcpTimeoutRehash   = "TcpTimeoutRehash"      // Total tcp timeout rehash
+	tcpDuplicateRehash = "TcpDupliateDataRehash" // Total tcp duplicates rehashed
+)
+
+var (
+	tcpTimeoutDesc = prometheus.NewDesc(
+		"tcp_timeout_rehash_count",
+		"Tcp timeout rehash count",
+		nil, nil,
+	)
+	tcpDuplicateDesc = prometheus.NewDesc(
+		"tcp_duplicate_rehash_count",
+		"Tcp duplicate rehash counts",
+		nil, nil,
+	)
+	segmentsReceivedDesc = prometheus.NewDesc(
+		"tcp_segments_received_count",
+		"TCP segments received on the node",
+		nil, nil,
+	)
+	segmentsSentDesc = prometheus.NewDesc(
+		"tcp_segments_sent_count",
+		"TCP segments sent from node",
+		nil, nil,
+	)
+	segmentsRetransmittedDesc = prometheus.NewDesc(
+		"tcp_segments_retransmitted_count",
+		"TCP segments retransmitted on the node",
+		nil, nil,
+	)
+)
+
+type kernelStatCollector struct {
+}
+
+func init() {
+	registerCollector("kernel_metrics", NewKernelStatCollector)
+}
+
+func NewKernelStatCollector() (Collector, error) {
+	return &kernelStatCollector{}, nil
+}
+
+// Parses a string which contains two lines, each starting with label
+// returning a map where the keys are the first lines elements, while the
+// second line contains the values, all casted to uint64, e.g
+//
+// label key1 key2 key3
+// label 1 2 3
+//
+// Outputs: {"key1":1 "key2":2 "key3":3}
+func parseKeyValueLines(output, label string) (map[string]uint64, error) {
+	stats := make(map[string]uint64)
+	splitFile := strings.Split(output, "\n")
+	for i, stat := range splitFile {
+		if !strings.Contains(stat, label) {
+			continue
+		}
+		keyList := strings.Fields(stat)
+		// TODO(#112): Look for next instance of label rather than assuming its the next
+		valList := strings.Fields(splitFile[i+1])
+		if len(valList) != len(keyList) {
+			// Return nothing if malformed.
+			return nil, fmt.Errorf("malformed key value pair while parsing %q", label)
+		}
+		for i, key := range keyList[1:] {
+			val, err := strconv.Atoi(valList[i+1])
+			if err != nil {
+				return nil, fmt.Errorf("could not cast value to uint64, got %q", err)
+			}
+			stats[key] = uint64(val)
+		}
+		return stats, nil
+	}
+	return stats, nil
+}
+
+func (c *kernelStatCollector) Update(ch chan<- prometheus.Metric) error {
+	// Get snmp values.
+	data, err := ioutil.ReadFile("/proc/net/snmp")
+	if err != nil {
+		return fmt.Errorf("could not read proc/net/snmp")
+	}
+	output := string(data)
+	netstats, err := parseKeyValueLines(output, snmpLabel)
+	if err != nil {
+		return err
+	}
+	if val, ok := netstats[tcpSegsIn]; ok {
+		ch <- prometheus.MustNewConstMetric(segmentsReceivedDesc, prometheus.CounterValue, float64(val))
+	}
+	if val, ok := netstats[tcpSegsOut]; ok {
+		ch <- prometheus.MustNewConstMetric(segmentsSentDesc, prometheus.CounterValue, float64(val))
+	}
+	if val, ok := netstats[tcpSegsRetrans]; ok {
+		ch <- prometheus.MustNewConstMetric(segmentsRetransmittedDesc, prometheus.CounterValue, float64(val))
+	}
+
+	// Get netstat values
+	data, err = ioutil.ReadFile("/proc/net/netstat")
+	if err != nil {
+		return fmt.Errorf("could not read proc/net/netstat")
+	}
+	output = string(data)
+	nstats, err := parseKeyValueLines(output, netstatLabel)
+	if err != nil {
+		return err
+	}
+	if val, ok := nstats[tcpTimeoutRehash]; ok {
+		ch <- prometheus.MustNewConstMetric(tcpTimeoutDesc, prometheus.CounterValue, float64(val))
+	}
+	if val, ok := nstats[tcpDuplicateRehash]; ok {
+		ch <- prometheus.MustNewConstMetric(tcpDuplicateDesc, prometheus.CounterValue, float64(val))
+	}
+	return nil
+}
--- a/pkg/metrics/collector/kernel_metrics_test.go
+++ b/pkg/metrics/collector/kernel_metrics_test.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2021 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package collector
+
+import (
+	"testing"
+)
+
+// Tests if got contains all of want.
+func mapContainsAll(got, want map[string]uint64) bool {
+	for k, v := range want {
+		if _, ok := got[k]; ok && v == got[k] {
+			continue
+		}
+		return false
+	}
+	return true
+}
+
+func TestParseNetstat(t *testing.T) {
+	t.Parallel()
+
+	fakeNetstat := `
+TcpExt: SyncookiesSent RcvPruned OfoPruned OutOfWindowIcmps LockDroppedIcmps TcpTimeoutRehash TcpDuplicateDataRehash TCPDSACKRecvSegs TCPDSACKIgnoredDubious
+TcpExt: 0 0 0 0 0 440 99 21800 3`
+	want := map[string]uint64{"TcpTimeoutRehash": 440, "TcpDuplicateDataRehash": 99}
+	got, err := parseKeyValueLines(fakeNetstat, netstatLabel)
+	if err != nil {
+		t.Errorf("failed parseNetstat, got %q", err)
+	}
+	if !mapContainsAll(got, want) {
+		t.Errorf("parseNetstatVals returns %+q, want %+q", got, want)
+	}
+}
+
+func TestParseSnmp(t *testing.T) {
+	t.Parallel()
+	fakeSnmp := `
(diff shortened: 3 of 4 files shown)


pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+		nil, nil,
+	)
+	segmentsReceivedDesc = prometheus.NewDesc(
+		"segments_received",
+		"Segments received on the node, collected from netstat -s",
+		nil, nil,
+	)
+	segmentsSentDesc = prometheus.NewDesc(
+		"segments_sent",
+		"Segments sent from node, collected from netstat -s",
```
varunmar, line 46: "TCP segments sent from node..."

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+		"Tcp duplicate rehash, collected from nstat",
+		nil, nil,
+	)
+	segmentsReceivedDesc = prometheus.NewDesc(
+		"segments_received",
+		"Segments received on the node, collected from netstat -s",
+		nil, nil,
+	)
+	segmentsSentDesc = prometheus.NewDesc(
+		"segments_sent",
```
varunmar, line 45: tcp_segments_sent

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+		"Tcp timeout rehash, collected from nstat",
+		nil, nil,
+	)
+	tcpDuplicateDesc = prometheus.NewDesc(
+		"tcp_duplicate_rehash",
+		"Tcp duplicate rehash, collected from nstat",
+		nil, nil,
+	)
+	segmentsReceivedDesc = prometheus.NewDesc(
+		"segments_received",
```
varunmar, line 40: tcp_segments_received

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+		"Segments received on the node, collected from netstat -s",
```
varunmar, line 41: "TCP segments received by the node..."

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+		nil, nil,
+	)
+	segmentsSentDesc = prometheus.NewDesc(
+		"segments_sent",
+		"Segments sent from node, collected from netstat -s",
+		nil, nil,
+	)
+	segmentsRetransmittedDesc = prometheus.NewDesc(
```
varunmar, line 49: Same deal here. Make it clear these are TCP metrics.

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+
+type kernelStats struct {
+	tcpTimeoutRehash      uint64
+	tcpDuplicateRehash    uint64
+	segmentsReceived      uint64
+	segmentsSent          uint64
+	segmentsRetransmitted uint64
+}
+
+func parseNetstatVals(stats *kernelStats, output string) error {
```
varunmar, line 75: Put an example line here as a comment, showing what you expect the input to look like.

Parsing tool output is always fraught, it's always worth it to make it clear what the input/output is, independently of how the parsing actually happens.

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		switch statList[2] {
+		case "received":
+			val, err := strconv.Atoi(statList[0])
+			if err != nil {
+				return fmt.Errorf("error casting to integer while getting segments received, err: %v", err)
+			}
+			statcount++
+			stats.segmentsReceived = uint64(val)
+		case "send":
```
varunmar, line 90: sent?

from net-tools v2.10-alpha, I see this TCP section which seems to have the right spellings..

Tcp:
    10124748 active connection openings
    6026748 passive connection openings
    11242 failed connection attempts
    7819340 connection resets received
    47 connections established
    502497809 segments received
    520382250 segments sent out
    15025 segments retransmitted
    7 bad segments received
    7847689 resets sent

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+			val, err := strconv.Atoi(statList[0])
+			if err != nil {
+				return fmt.Errorf("error casting to integer while getting segments sent, err: %v", err)
+			}
+			statcount++
+			stats.segmentsSent = uint64(val)
+		case "retransmited":
```
varunmar, line 97: here too -
the 2.10-alpha version seems to have spelled it right.

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+			val, err := strconv.Atoi(statList[0])
+			if err != nil {
+				return fmt.Errorf("error casting to integer while getting segments received, err: %v", err)
+			}
+			statcount++
+			stats.segmentsReceived = uint64(val)
+		case "send":
+			val, err := strconv.Atoi(statList[0])
+			if err != nil {
+				return fmt.Errorf("error casting to integer while getting segments sent, err: %v", err)
```
varunmar, line 93: consider not returning early here if we can't parse it - keep going and see if we can parse anything at all.

pkg/metrics/collector/kernel_metrics.go
varunmar, line 75: Consider returning the kernelStats struct as output.

I'm actually not sure why you have it here as a param, let me know if I am missing something.

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+			stats.segmentsRetransmitted = uint64(val)
+		}
+	}
+	if statcount != 3 {
+		return fmt.Errorf("could not get all stats for netstat, got %d/3, %+v", statcount, stats)
+	}
+	return nil
+}
+
+func parseNstatVals(stats *kernelStats, output string) error {
```
varunmar, line 112: Same thing with this function too - add an example parse line, consider not taking the struct as input, do not return early if there is a parsing failure.

varunmar (changes requested):
Make sure the commits are squashed before sending them out for review - in this case, the entire change can be a single commit.

pkg/metrics/collector/kernel_metrics.go
michaelasp: Strange I also have net-tools 2.10-alpha, but running it on alpine seems to give this output:

Tcp:
    0 active connections openings
    0 passive connection openings
    0 failed connection attempts
    0 connection resets received
    0 connections established
    0 segments received
    0 segments send out
    0 segments retransmited
    0 bad segments received.
    0 resets sent

Try running the netd container with net-tools installed and see what you get.

pkg/metrics/collector/kernel_metrics.go
michaelasp: Same as above, same version with different outputs, strange.

pkg/metrics/collector/kernel_metrics.go
michaelasp: Main reason I had it as a param was because there are 2 functions which fill this struct. However I think it migh be better to declare two structs, one for nstat and one for netstat. Let me know what you think.

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Sg!

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	tcpTimeoutDesc = prometheus.NewDesc(
+		"tcp_timeout_rehash_count",
+		"Tcp timeout rehash, collected from nstat",
```
varunmar, line 30: No need to mention the source of this data here. (Also, it's already outdated...)

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+func parseNetstatVals(output string) map[string]uint64 {
+	stats := make(map[string]uint64)
+	splitFile := strings.Split(output, "\n")
+	for i, stat := range splitFile {
+
+		if !strings.Contains(stat, "TcpExt:") {
+			continue
+		}
+		keyList := strings.Fields(stat)
+		valList := strings.Fields(splitFile[i+1])
```
varunmar, line 83: Add a TODO here to look for the next instance of TcpExt: instead of assuming it's the immediately following line.

No need to change it, just leave a TODO here as a breadcrumb if anybody comes looking later.

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+// parseSnmpVals parses the contents of /proc/net/snmp
+// looking for lines such as:
+//
+// Tcp: RtoAlgorithm RtoMin RtoMax MaxConn ActiveOpens PassiveOpens AttemptFails EstabResets CurrEstab InSegs OutSegs RetransSegs InErrs OutRsts InCsumErrors
+// Tcp: 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
+//
+// and populates a map with all the parsed values where the key are the labels in the first line separated by spaces
+// and the values are the numbers in the second line, casted to uint64
+
+func parseSnmpVals(output string) map[string]uint64 {
```
varunmar, line 105: This looks like exactly the same function as above. Maybe add a helper that takes in the two lines and returns the map.

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	return stats
+}
+
+func (c *kernelStatCollector) Update(ch chan<- prometheus.Metric) error {
+	// Get snmp values.
+	data, err := ioutil.ReadFile("/proc/net/snmp")
+	if err == nil {
+		output := string(data)
+		netstats := parseSnmpVals(output)
+		if val, ok := netstats["InSegs"]; ok {
```
varunmar, line 132: Pull these keys up into package consts, and give them a comment each.

pkg/metrics/collector/kernel_metrics_test.go
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+package collector
+
+import (
+	"testing"
+)
+
+// Tests if a contains all of b.
+func mapContainsAll(a, b map[string]uint64) bool {
```
varunmar, line 24: nit: consider giving them easier to remember names?
"needle" and "haystack" is common when looking for single items in a collection.
Here, maybe just "got" and "want" is appropriate?

varunmar (commented):
Very nice! A couple minor comments, and a little bit of de-duplication of one function.

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done, have a function that takes the label which is used in the update function

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+		if val, ok := netstats[TCPSegsRetrans]; ok {
+			ch <- prometheus.MustNewConstMetric(segmentsRetransmittedDesc, prometheus.CounterValue, float64(val))
+		}
+
+	}
+
+	// Get netstat values
+	data, err = ioutil.ReadFile("/proc/net/netstat")
+	if err != nil {
+		return nil
```
MrHohn, line 134: Here and the above - let's return the error?

MrHohn: I do see in `conntrack.go` we ignore a similar error and return `nil` but I don't think we should hide it :/ I will fix that separately.

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		return stats
+	}
+	return stats
+}
+
+func (c *kernelStatCollector) Update(ch chan<- prometheus.Metric) error {
+	// Get snmp values.
+	data, err := ioutil.ReadFile("/proc/net/snmp")
+	if err == nil {
```
MrHohn, line 116: Would be great to keep the normal code path unindented: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

I think it is fine to just return the error if `ReadFile` failed - the chance that it fails on `/proc/net/snmp` but suceeds on `/proc/net/netstat` should be low anyway.

MrHohn (commented):
Thanks for adding these - generally looks good to me.

pkg/metrics/collector/kernel_metrics.go
michaelasp: SG!

pkg/metrics/collector/kernel_metrics.go
michaelasp: Sounds good!

pkg/metrics/collector/kernel_metrics_test.go
```diff
@@ -0,0 +1,62 @@
+/*
+Copyright 2018 Google Inc.
```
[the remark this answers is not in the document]
michaelasp: Done

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+	stats := make(map[string]uint64)
+	splitFile := strings.Split(output, "\n")
+	for i, stat := range splitFile {
+
+		if !strings.Contains(stat, label) {
+			continue
+		}
+		keyList := strings.Fields(stat)
+		// TODO: Look for next instance of label rather than assuming its the next
+		valList := strings.Fields(splitFile[i+1])
```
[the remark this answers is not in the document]
michaelasp: Done, also made it return an error since this is a bad path

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+		for i, key := range keyList[1:] {
+			val, err := strconv.Atoi(valList[i+1])
+			if err != nil {
```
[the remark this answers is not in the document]
michaelasp: Made an error since this is a faulty path and should be investigated

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+// Keys for metrics after parsing
+const (
+	// Got from Snmp counters
+	TCPSegsIn      = "InSegs"      // Total tcp segments in
+	TCPSegsOut     = "OutSegs"     // Total tcp segments out
+	TCPSegsRetrans = "RetransSegs" // Total tcp segments retransmitted
+	// Got from netstat
+	TCPTimeoutRehash   = "TcpTimeoutRehash"      // Total tcp timeout rehash
+	TCPDuplicateRehash = "TcpDupliateDataRehash" // Total tcp duplicates rehashed
+)
```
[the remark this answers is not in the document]
michaelasp: Nope, good catch. Fixed!

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,145 @@
+/*
+Copyright 2018 Google Inc.
```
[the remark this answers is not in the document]
michaelasp: Done

netd.yaml
```diff
@@ -108,7 +108,7 @@ metadata:
   labels:
     addonmanager.kubernetes.io/mode: EnsureExists
 data:
-  metrics_collectors: "conntrack,socket"
+  metrics_collectors: "conntrack,socket,kernel_metrics"
```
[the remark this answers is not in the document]
michaelasp: SG

pkg/metrics/collector/kernel_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+func parseKeyValueLines(output, label string) map[string]uint64 {
+	stats := make(map[string]uint64)
+	splitFile := strings.Split(output, "\n")
+	for i, stat := range splitFile {
+
+		if !strings.Contains(stat, label) {
+			continue
+		}
+		keyList := strings.Fields(stat)
+		// TODO: Look for next instance of label rather than assuming its the next
```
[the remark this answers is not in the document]
michaelasp: Done

MrHohn (approved):
Thanks! Will leave the final LGTM to Varun :)
/approve

PR #111, opened 2021-07-27
Add netlink framework and example metric

michaelasp:
Using netlink for fine grained metrics without any external dependencies besides netlink which is already in netd. Updated sys and netlink packages in the process. Currently a proof of concept metric which will be extended in the future.


varunmar:
Here, it makes sense to separate the go.mod + vendor changes (just for updating the netlink version) from the real code that adds the netlink framework.
The idea is to separate each commit into something meaningful and standalone that makes sense when reading the commit log on master after merging.

michaelasp:
Made it 2 commits rather than 1

.golangci.yml
```diff
@@ -27,6 +27,9 @@ run:
   # on Windows.
   skip-dirs:
     - .go
+    - pkg/inetdiag
```
varunmar, line 30: Why exclude these?

varunmar (commented):
I think we should slightly change the directory structure here - both to organize the code and to satisfy the attribution requirements from the apache 2.0 license.

Let's create a pkg/tpc_metrics directory, where we can add a NOTICE file with the following content:

Copyright 2021 Google Inc.

Portions of this directory derived from https://github.com/m-lab/tcp-info under the terms of the Apache License, Version 2.0

.golangci.yml
michaelasp: A lot of the enum names are fully capitalized, since they are meant to mimic the Linux kernel enums. The linter doesn't like that and refuses to build with them as is. If there's a better way of doing this please let me know!

varunmar:
LGTM from my perspective. I'll let Zihong review and give the final approval

pkg/tcp_metrics/parser/parser_linux.go
```diff
@@ -0,0 +1,48 @@
+// Package netlink contains the bare minimum needed to partially parse netlink messages.
```
MrHohn, line 1: nit: Package parser contains...

pkg/tcp_metrics/parser/parse.go
```diff
@@ -0,0 +1,321 @@
...
+			return nil, err
+		}
+		ra := NetlinkRouteAttr{Attr: RtAttr(*a), Value: vbuf[:int(a.Len)-SizeofRtAttr]}
+		attrs = append(attrs, ra)
+		b = b[alen:]
+	}
+	return attrs, nil
+}
+
+// MakeArchivalRecord parses the NetlinkMessage into a ArchivalRecord.  If skipLocal is true, it will return nil for
```
MrHohn, line 77: nit: MakeSnapShot...

MrHohn (approved):
Thanks for adding these - I saw there are a big ton of info available from this netlink framework and this seems to be good start!

I have only nit comments as many of the netlink codes are kept intact. Meanwhile approving to unblock the other PR - feel free to fix them in a follow up.

/lgtm
/approve

PR #113, opened 2021-08-09
Per pod metrics

michaelasp:
Per pod metrics, creates a watch on pods for that specific node in order to filter the watch. Node name is given through the environment variable CURRENT_NODE_NAME. Once a watch is established we scan through all namespaces, grabbing all tcp connections and reporting it if it matches a pod ip. 

We accumulate retransmits and active connections but this can be augmented with any statistic from the snapshot struct in pkg/tcp_metrics/parser/parser.go


pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,29 +15,156 @@ package collector
...
+	if err != nil {
+		return err
+	}
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+	kubeInformerFactory := informers.NewSharedInformerFactory(clientset, time.Second*30)
+	podInformer := kubeInformerFactory.Core().V1().Pods().Informer()
```
varunmar, line 110: To restrict this to just pods on the current node, add a field selector with spec.nodeName=<node_name>

See for example: https://github.com/cilium/cilium/blob/master/pkg/k8s/watchers/pod.go#L123

(They use ParseSelectorOrDie, which is a bit extreme for us here. No need to die here, this watch is not critical to the operation of netd. Use one of the other functions; like OneTermEqualSelector in  https://github.com/kubernetes/apimachinery/blob/master/pkg/fields/selector.go#L465

michaelasp: Done

varunmar:
/assign mrhohn

This PR follows up on the one for adding the netlink framework.

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,287 @@ package collector
...
+	mapMux.RLock()
+	defer mapMux.RUnlock()
+	podDef, ok := ipMap[key]
+	return podDef, ok
+}
+
+// Functionality for pod watching
+func createPodWatch() error {
+	// creates the in-cluster config
+	config, err := rest.InClusterConfig()
```
MrHohn, line 107: I originally wanted to say that this may make kube-proxy/dataplane v2 a new dependency of netd because this `InClusterConfig` uses `default/kubernetes` service's cluster IP by default. Then I realized we intentionally override `KUBERNETES_SERVICE_HOST` to the actual master IP during the deployment. So this should just work :)

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,287 @@ package collector
...
+
+type podDef struct {
+	name      string
+	namespace string
+}
+
+var (
+	mapMux   sync.RWMutex
+	ipMap    = make(map[string]podDef)
+	nodeName = os.Getenv("CURRENT_NODE_NAME")
```
MrHohn, line 70: Interesting - wasn't aware of this env. Do you have a link to the documentation? Or do you plan to inject this from the manifest using `spec.nodeName`?

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,287 @@ package collector
...
+	}
+	return statMap
+}
+
+func NewNetlinkCollector() (Collector, error) {
+	return &netlinkCollector{}, nil
+}
+
+func getPids() ([]int, error) {
+	files, err := ioutil.ReadDir("/proc")
```
MrHohn, line 211: Humm, I honestly don't know - would the actual `/proc` folder from the host be shared with the netd pod? Do we need to add any escalated privilege for this?

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,287 @@ package collector
...
+	pod := podObj.(*v1.Pod)
+	if podAllow(pod) {
+		safeIPWrite(pod.Status.PodIP, podDef{name: pod.ObjectMeta.Name, namespace: pod.ObjectMeta.Namespace})
+	}
+}
+
+func onPodUpdate(oldPod, newPod interface{}) {
+	prevPod := oldPod.(*v1.Pod)
+	pod := newPod.(*v1.Pod)
+	if prevPod.Status.PodIP == pod.Status.PodIP {
```
MrHohn, line 162: I'm under the impression that pod IP is immutable so we don't need not to watch for pod updates?

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -108,25 +310,19 @@ func getRequests() ([]*parser.Snapshot, error) {
...
-
 func (c *netlinkCollector) Update(ch chan<- prometheus.Metric) error {
 	snapshots, err := getRequests()
 	if err != nil {
+		glog.Infof("Could not get requests")
 		return fmt.Errorf("could not get tcp requests: %q", err)
 	}
-	retransmits := getRetransmits(snapshots)
-	ch <- prometheus.MustNewConstMetric(TCPConnectionsDesc, prometheus.GaugeValue, float64(len(snapshots)))
-	ch <- prometheus.MustNewConstMetric(TCPConnectionsDesc, prometheus.GaugeValue, float64(retransmits))
```
MrHohn: I think it is still useful to have separate node-level metrics in additional to the pod metrics. Could we keep these?

pkg/metrics/collector/netlink_metrics_test.go
```diff
@@ -0,0 +1,76 @@
+/*
+Copyright 2018 Google Inc.
```
MrHohn, line 2: nit: Copyright 2021 Google Inc.

MrHohn (commented):
Thanks for adding these! Added a few comments for the first pass.

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,287 @@ package collector
...
+	glog.Infof("Initializing netlink metrics")
+	err := createPodWatch()
+	if err != nil {
+		glog.Infof("Error, pod watch not running with err, %q", err)
+	}
+
+	registerCollector("netlink_metrics", NewNetlinkCollector)
+}
+
+// Functionality for multithreaded map usage
```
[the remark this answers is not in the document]
michaelasp: Yup, will do!

pkg/metrics/collector/netlink_metrics.go
michaelasp: When a pod gets created the pod ip is sometimes not initialized yet. Once the pod is updated to ready, the ip gets added so there are a few cases where we need to watch for pod updates. I ran a test cluster with this configuration and it seemed to confirm this.

pkg/metrics/collector/netlink_metrics.go
michaelasp: Yeah, unfortunately we need to have the namespaces of the other containers on the node which requires access to their proc folders. This requires hostPID to be set to true in the netd pod spec. Without it we can only read our namespace since containerd/docker don't expose the namespaces of other containers in the cluster. Let me know what you think about this.

pkg/metrics/collector/netlink_metrics.go
michaelasp: Our current plan is to inject this through the manifest, I can give you the example internally.

pkg/metrics/collector/netlink_metrics.go
michaelasp: Cool!

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,287 @@ package collector
...
+var (
+	mapMux   sync.RWMutex
+	ipMap    = make(map[string]podDef)
+	nodeName = os.Getenv("CURRENT_NODE_NAME")
+)
+
+func init() {
+
+	glog.Infof("Initializing netlink metrics")
+	err := createPodWatch()
```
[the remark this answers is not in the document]
michaelasp: Yup, my thoughts were to make it so this collector would only be enabled when we want fine grained metrics but since we also want the more aggregate metrics we can do this!

pkg/metrics/collector/netlink_metrics.go
michaelasp: Done!

pkg/metrics/collector/netlink_metrics_test.go
michaelasp: Done

pkg/metrics/collector/netlink_metrics.go
michaelasp: Sg!

pkg/metrics/collector/netlink_metrics.go
varunmar: Yes, we have to watch for pod updates. Kubelet only starts the CRI/CNI after the pod object exists.

pkg/metrics/collector/netlink_metrics.go
varunmar: +1 - we added that service host IP because the init container also makes a call to the master (to get the current node's pod cidr and write the ipam section of the CNI config file)

pkg/metrics/collector/netlink_metrics.go
varunmar: Good point.
Michael, when you add these back can you also add a comment that the difference between these metrics vs a sum over all pod metrics is that these will include metrics from host network pods + processes on the root ns as well.

michaelasp: Sounds good!

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,293 @@ package collector
...
 
+type netlinkCollector struct {
+}
+
+type netlinkStats struct {
+	retransmits    uint64
+	tcpConnections uint64
+}
+
+type podDef struct {
```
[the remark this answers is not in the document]
michaelasp: That's true, removed parsing and just kept the pod object unchanged

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,293 @@ package collector
...
-		"tcp_connections",
+		"active_tcp_connections",
 		"Current amount of tcp connections on a node",
-		nil, nil,
+		[]string{"pod_name", "namespace_name"}, nil,
 	)
 
 	ActiveTCPRetransmits = prometheus.NewDesc(
 		"active_tcp_retransmits",
 		"Current amount of tcp retransmits on a node",
```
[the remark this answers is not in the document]
michaelasp: Done

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,293 @@ package collector
...
+	"github.com/GoogleCloudPlatform/netd/pkg/tcp_metrics/inetdiag"
+	"github.com/GoogleCloudPlatform/netd/pkg/tcp_metrics/parser"
+	"github.com/GoogleCloudPlatform/netd/pkg/tcp_metrics/tcp"
 )
 
 var (
 	TCPConnectionsDesc = prometheus.NewDesc(
-		"tcp_connections",
+		"active_tcp_connections",
 		"Current amount of tcp connections on a node",
```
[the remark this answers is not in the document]
michaelasp: Done

pkg/metrics/collector/netlink_metrics.go
MrHohn: Great, thanks for keeping this.

MrHohn (commented):
Thanks - I essentially have one comment on the netns approach and the rest looks good to me.

netd.yaml
```diff
@@ -233,3 +247,7 @@ spec:
         hostPath:
           path: /home/kubernetes/bin
           type: Directory
+      - name: net-ns
+        hostPath:
+          path: /var/run/netns
```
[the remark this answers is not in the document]
michaelasp: Unfortunately seems like we need to use hostpid, seems like the files don't get mounted properly so we only get the namespaces at the initialization of netd. HostPID is far more reliable. I don't like exposing the pids either but it seems like that's the only way forward to get reliable metrics.

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,293 @@ package collector
...
+				if err != nil {
+					return nil, err
+				}
+				if m != nil {
+					cur, err := parser.MakeSnapShot(m, true)
+
+					if cur == nil || err != nil {
+						continue
+					}
+					if val, ok := cookieSet[cur.SockInfo.Cookie]; ok {
```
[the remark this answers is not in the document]
michaelasp: Nope, removed that part

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,293 @@ package collector
...
+	)
+
+	stopper := make(chan struct{})
+
+	go controller.Run(stopper)
+
+	return nil
+}
+
+func podAllow(pod *v1.Pod) bool {
```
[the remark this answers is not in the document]
michaelasp: Done

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,293 @@ package collector
...
+	tcpConnections uint64
+}
+
+type podDef struct {
+	name      string
+	namespace string
+}
+
+var (
+	mapMux   sync.RWMutex
```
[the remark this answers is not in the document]
michaelasp: Done

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,293 @@ package collector
...
+	glog.Infof("Initializing netlink metrics")
+	err := createPodWatch()
+	if err != nil {
+		glog.Infof("Error, pod watch not running with err, %q", err)
+	}
+
+	registerCollector("netlink_metrics", NewNetlinkCollector)
+}
+
+// Functionality for multithreaded map usage
```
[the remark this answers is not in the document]
michaelasp: Good point, especially for extensibility, done!

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -15,85 +15,293 @@ package collector
...
+		}
+	}
+	return statMap
+}
+
+func NewNetlinkCollector() (Collector, error) {
+	return &netlinkCollector{}, nil
+}
+
+func getPids() ([]int, error) {
```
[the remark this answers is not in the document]
michaelasp: Done

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -108,25 +316,19 @@ func getRequests() ([]*parser.Snapshot, error) {
...
-		}
-		retransmits += uint64(snap.TCPInfo.Retrans)
-	}
-	return retransmits
-}
-
 func (c *netlinkCollector) Update(ch chan<- prometheus.Metric) error {
 	snapshots, err := getRequests()
 	if err != nil {
+		glog.Infof("Could not get requests")
```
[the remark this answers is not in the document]
michaelasp: Nope, removed

pkg/metrics/collector/netlink_metrics_test.go
```diff
@@ -0,0 +1,76 @@
...
+	if statMap[myPod] != want[myPod] {
+		t.Fatalf("Fatal, got %+q, wanted %+q", statMap, want)
+	}
+}
+
+func TestSafeIPAdd(t *testing.T) {
+	myPod := podDef{name: "pod", namespace: "default"}
+	safeIPWrite("my-ip", myPod)
+	if ipMap["my-ip"] != myPod {
+		t.Fatalf("Fatal, got %+q, wanted %+q", ipMap["my-ip"], myPod)
```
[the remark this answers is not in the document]
michaelasp: True, will change

netd.yaml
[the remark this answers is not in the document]
MrHohn: Okay, in that case, lets go with the pid approach for now. We will probably keep the pod watch portion disabled by default until we can properly lock down the netd image. WDYT?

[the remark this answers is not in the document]
varunmar: SGTM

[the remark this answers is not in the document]
michaelasp: Sounds good! These metrics should be able to be enabled at any time.

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -14,84 +16,318 @@ limitations under the License.
...
 	}
-	// Adapted this from req.Execute in nl_linux.go
-snapshotLoop:
-	for {
-		msgs, _, err := s.Receive()
+	for _, curNs := range namespaces {
+		defer curNs.Close()
+		s, err := nl.SubscribeAt(curNs, basens, sockType)
 		if err != nil {
+			continue
```
MrHohn, line 290: Could we log the error?

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -121,12 +357,33 @@ func getRetransmits(snapshots []*parser.Snapshot) uint64 {
 }
 
 func (c *netlinkCollector) Update(ch chan<- prometheus.Metric) error {
+	if enablePodWatch && firstRun {
+		err := createPodWatch()
+		if err != nil {
```
MrHohn, line 362: nit: make it a one liner:
```
if err := createPodWatch(); err != nil {
    ...
```

pkg/metrics/collector/netlink_metrics.go
```diff
@@ -14,84 +16,318 @@ limitations under the License.
...
+		}
+	}
+	return statMap
+}
+
+func NewNetlinkCollector() (Collector, error) {
+	return &netlinkCollector{}, nil
+}
+
+func getNamespaces() ([]netns.NsHandle, error) {
```
MrHohn, line 237: Just double checking - if we disable pod watch and don't enable HostPID fo netd, would it still be able to expose the node level metrics properly (I guess yes since we will be extracting the host network namespace from netd its own's PID)?

MrHohn (approved):
Thanks Micael, I have a few final comments.

pkg/metrics/collector/netlink_metrics.go
michaelasp: Hmm, there may be some issues because we may not count all connections on every namespace. There may be some undercounting involved.

pkg/metrics/collector/netlink_metrics.go
michaelasp: Done

pkg/metrics/collector/netlink_metrics.go
michaelasp: Yup!

pkg/metrics/collector/netlink_metrics.go
MrHohn: Got it - do you mind filing an issue so that we don't lost track of this? Then we could merge this PR.

varunmar: Hmm. I think we can do two things - always collect metrics for the root ns (and label it by the node IP I guess) and optionally if the pod watch is enabled, also collect metrics for the pod namespaces.

I think this will require some additional work to figure out how to export the root ns metrics into a K8sContainer schema. So for now, I think just create an issue so we don't lose track and we can merge this PR.

varunmar (approved):

PR #117, opened 2022-01-17
Update dependencies and install-cni script to work with debian-buster

varunmar:
The install script uses bash-isms, and requires the use of the "ip" command to find the MTU

Install these dependencies and modify the shebang line to use /bin/bash instead of /bin/sh


MrHohn (approved):
Thanks for catching this Varun. Let me make sure this fix can catch the next release.
/lgtm
/approve

PR #118, opened 2022-01-25
Add Makefile targets to push multi-arch container manifests

tstapler:
The docker images published for netd releases are not [multi arch images](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) (docker images that support multiple CPU architectures). Currently the Makefile is configured to publish a separate image tag for each architecture. Images are tagged using the format `netd-{arch}:version` for example `us.gcr.io/gke-release-staging/netd-amd64:v0.3.0-gke.0` and . The target CPU architectures are specified in the [Makefile](https://github.com/GoogleCloudPlatform/netd/blob/v0.3.0/Makefile#L43). 

What is missing is a multi arch manifest. After this change, when a maintainer (or the ci/cd system) runs `make all-push` new image `gcr.io/gke-release-staging/netd` should be created with a multi arch image manifest. The image will contain references to the architecture specific images, and downstream consumers can just use `gcr.io/gke-release-staging/netd` and their container engine will pull the correct image at runtime.

These changes should not change the workflow for publishing new netd images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ ALL_ARCH := amd64 arm arm64 ppc64le
 
 BASE_IMAGE ?= k8s.gcr.io/build-image/debian-iptables-$(ARCH):buster-v1.8.0
 
+MANIFEST_IMAGE := $(REGISTRY)/$(BIN)
 IMAGE := $(REGISTRY)/$(BIN)-$(ARCH)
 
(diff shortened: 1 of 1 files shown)


MrHohn (approved):
Thanks for adding this Tyler. I didn't spot obvious issue and have tested this locally, which seems to work.

/lgtm
/approve

Issue #119, opened 2022-03-10
Bash-specific syntax in install-cni.sh


jingyuanliang:
https://github.com/GoogleCloudPlatform/netd/blob/master/scripts/install-cni.sh declares shebang line `#!/bin/bash` and sporadically uses bash-specific syntax not available in POSIX sh.

https://github.com/GoogleCloudPlatform/netd/blob/master/netd.yaml calls it using `command: ["sh", "/install-cni.sh"]` which doesn't work as expected when this `sh` is not bash.

I don't know why it works at startup (I guess it's running directly there called by someone else so the shebang line takes effect, but I didn't trace to the place where it's started at startup), but when netd is being restarted, this begins to hurt. I captured the following error messages ("Not using Cilium plug-in." is from my uncommitted changes and it actually shouldn't go into that branch) from `kubectl logs netd-m5277 -n kube-system -c install-cni`:

```
/install-cni.sh: 23: /install-cni.sh: [[: not found
/install-cni.sh: 42: [: false: unexpected operator
Calico network policy config:  false
Not using Cilium plug-in.
/install-cni.sh: 64: [: true: unexpected operator
/install-cni.sh: 71: [: true: unexpected operator
PodCIDR validation succeeded: "10.76.0.0/24"
Filling IPv4 subnet "10.76.0.0/24"
/install-cni.sh: 105: [: true: unexpected operator
/install-cni.sh: 133: [: false: unexpected operator
/install-cni.sh: 133: [: false: unexpected operator
Clearing IPv6 subnet and route given private IPv6 access is disabled...
Set the default mtu to 1460, inherited from dev eth0
Creating CNI spec...
```

jingyuanliang:
It seems this was broken since https://github.com/GoogleCloudPlatform/netd/commit/195dd49aded9645f3699232c652243d25d219d2c

PR #121, opened 2022-03-10
Wait until cilium-cni binary is ready when ENABLE_CILIUM_PLUGIN is set

jingyuanliang:
I'm avoiding bash-specific syntax ([[ ]] instead of [ ]) due to #119 (even if I'm fixing the configuration file to use bash instead of sh, I'm not sure about the direction of the next step, so better to be safe here).

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -71,6 +71,12 @@ fi
 if [ "${ENABLE_CILIUM_PLUGIN}" == "true" ]
 then
   cni_spec=$(echo ${cni_spec:-} | sed -e "s#@cniCiliumPlugin#,{\"type\": \"cilium-cni\"}#g")
+  cilium_bin="/host/home/kubernetes/bin/cilium-cni"
+  while [ ! -f "$cilium_bin" ]
+  do
+    echo "Waiting for $cilium_bin to be ready..."
+    sleep 2
+  done
 else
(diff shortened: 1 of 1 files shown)


scripts/install-cni.sh
```diff
@@ -56,23 +56,32 @@ fi
...
+  cni_spec=$(echo ${cni_spec:-} | sed -e "s#@cniBandwidthPlugin#,{\"type\": \"bandwidth\",\"capabilities\": {\"bandwidth\": true}}#g")
 else
-	cni_spec=$(echo ${cni_spec:-} | sed -e "s#@cniBandwidthPlugin##g")
+  cni_spec=$(echo ${cni_spec:-} | sed -e "s#@cniBandwidthPlugin##g")
 fi
 
 if [ "${ENABLE_CILIUM_PLUGIN}" == "true" ]
 then
-        cni_spec=$(echo ${cni_spec:-} | sed -e "s#@cniCiliumPlugin#,{\"type\": \"cilium-cni\"}#g")
+  echo "Adding Cilium plug-in:"
```
MrHohn, line 73: nit: "Adding Cilium plug-in to the CNI config." to be clear?

MrHohn (approved):
Thanks Jingyuan, LGTM with a minor comment.
/approve

scripts/install-cni.sh
jingyuanliang: Done

MrHohn (approved):
/lgtm

PR #122, opened 2022-04-07
Add pod_ip_metrics collector to netd

sdmodi:
This new module emits metrics for pod IP allocation by examining the gke-pod-network directory

### DIFF
--- a/pkg/metrics/collector/pod_ip_metrics.go
+++ b/pkg/metrics/collector/pod_ip_metrics.go
@@ -1,0 +1,179 @@
+/*
+Copyright 2022 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package collector
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"os"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type ipFamily int
+
+const (
+	gkePodNetworkDir = "/host/var/lib/cni/networks/gke-pod-network"
+	ipv4 ipFamily = iota
+	ipv6 = 1
+	dual = 2
+	dualStack = "IPV4_IPV6"
+)
+
+var (
+	usedIpv4AddrCountDesc = prometheus.NewDesc(
+		"used_ipv4_addr_count",
+		"Indicates how many IPv4 addresses are in use.",
+		nil, nil,
+	)
+	usedIpv6AddrCountDesc = prometheus.NewDesc(
+		"used_ipv6_addr_count",
+		"Indicates how many IPv6 addresses are in use.",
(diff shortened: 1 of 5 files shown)


pkg/metrics/collector/pod_ip_metrics_test.go
```diff
@@ -0,0 +1,132 @@
...
+	defer f.Close()
+	if _, err := f.WriteString(content); err != nil {
+		t.Fatalf("Error writing to file %s: %v", path, err)
+	}
+}
+
+func mustCreateIpAddrDir(t *testing.T, dir string) string {
+	tempDir, err := ioutil.TempDir("", dir)
+	if err != nil {
+		t.Fatalf("error creating dir %s: %v", dir, err)
```
MrHohn, line 24: nit: error -> Error

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -0,0 +1,174 @@
...
+	c.usedIpv6AddrCount = ipv6Count
+	c.dualStackCount = dualCount
+	c.dualStackErrorCount = dualErrCount
+	c.duplicateIpCount = dupIpCount
+	return nil
+}
+
+func (c *podIpMetricsCollector) Update(ch chan<- prometheus.Metric) error {
+	if err := c.listIpAddresses(gkePodNetworkDir); err != nil {
+		return nil
```
MrHohn, line 165: Humm, probably log the error here instead? I know listIpAddresses() already logs that - but at the first glance it seems like omitting the error without looking into that func.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -0,0 +1,174 @@
...
+
+	s := bufio.NewScanner(buf)
+	s.Scan()
+	return s.Text(), s.Err()
+}
+
+func (c *podIpMetricsCollector) listIpAddresses(dir string) error {
+	f, err := os.Open(dir)
+	if err != nil {
+		glog.Errorf("Error opening directory %s, %v", dir, err)
```
MrHohn, line 106: Just wanted to point out there may be a short period of time when the `gke-pod-network` folder doesn't exist yet (e.g. no pod has been created or only hostnetwork pods are created). That however should be fine as we only log an error and retry later.

MrHohn (approved):
Thanks Sudeep, seems straight forward. Just a few nits - this LGTM.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -0,0 +1,174 @@
...
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type ipFamily int
+
+const (
+	gkePodNetworkDir = "/host/var/lib/cni/networks/gke-pod-network"
+	ipv4 ipFamily = 0
```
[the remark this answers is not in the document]
sdmodi: Not sure what you mean. I did add a String function for this type.

pkg/metrics/collector/pod_ip_metrics.go
sdmodi: Yes. I noticed that and this is ok.

pkg/metrics/collector/pod_ip_metrics.go
sdmodi: Done

MrHohn (approved):
/lgtm
/approve

pkg/metrics/collector/pod_ip_metrics.go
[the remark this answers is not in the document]
MrHohn: Yep changing to use iota is what I meant. Thx.

PR #123, opened 2022-04-12
Add fsnotify to watch files in the gke-pod-network directory

sdmodi:
By doing this, we will not parse the entire gke-pod-network directory every 30 seconds. Instead, we will only parse the directory every time there is a change in the pod IP allocation. For most cases, this will be more efficient.


pkg/metrics/collector/pod_ip_metrics_test.go
```diff
@@ -18,6 +19,14 @@ func mustCreateFile(t *testing.T, dir, name, content string) {
 	}
 }
 
+func mustDeleteFile(t *testing.T, dir, name string) {
+	path := dir + "/" + name
+	err := os.Remove(path)
+	if err != nil {
```
sypakine, line 25: nit: if err := os.Remove(path); err != nil {

pkg/metrics/collector/pod_ip_metrics_test.go
```diff
@@ -130,3 +139,75 @@ func TestListIpAddresses(t *testing.T) {
...
+		t.Errorf("dualStackCount. want: 2, got %d", mc.dualStackCount)
+	}
+	if mc.dualStackErrorCount != 0 {
+		t.Errorf("dualStackErrorCount. want: 0, got %d", mc.dualStackErrorCount)
+	}
+	if mc.duplicateIpCount != 0 {
+		t.Errorf("duplicateIpCount. want: 0, got %d", mc.duplicateIpCount)
+	}
+	if !podIpMetricsWatcherSetup {
+		t.Errorf("podIpMetricsWatcherSetup is not set to true")
```
sypakine, line 173: nit: reads a bit awkward, maybe use the same convention as the rest? 

podIpMetricsWatcherSetup. want: true, got: false

sypakine, line 173: Should this be Fatal?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -23,18 +23,19 @@ import (
...
 
 type ipFamily int
 
 const (
-	gkePodNetworkDir = "/host/var/lib/cni/networks/gke-pod-network"
-	ipv4 ipFamily = iota
-	ipv6 = 1
-	dual = 2
-	dualStack = "IPV4_IPV6"
+	gkePodNetworkDir          = "/host/var/lib/cni/networks/gke-pod-network"
```
sypakine, line 34: May as well update to the following to take advantage of iota.

const (
	ipv4 ipFamily = iota
	ipv6
	dual

	gkePodNetworkDir = "/host/var/lib/cni/networks/gke-pod-network"
	dualStack                = "IPV4_IPV6"
)

sypakine (approved):
LGTM with some small nits.

PR #135, opened 2022-06-24
install-cni: check DPV2_MIGRATION_READY when MIGRATE_TO_DPV2 is set

jingyuanliang:
When MIGRATE_TO_DPV2 is true and node label "cloud.google.com/gke-dpv2-migration-ready"
is not true, override ENABLE_CILIUM_PLUGIN to become false.

/assign MrHohn
/cc weil0ng
/hold


jingyuanliang:
cc @Weil0ng

MrHohn (approved):
Thanks for getting this out so quickly!
/lgtm
/approve

PR #137, opened 2022-07-21
`go mod tidy` and `go mod vendor`

jingyuanliang:
Note: do not tag after this is merged. Let's confirm if Louhi builds or not first.

If Louhi builds, I'll improve Louhi flow/GCB then tag here.

/assign @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #139, opened 2022-07-22
Fix build.sh for netd-init to use $GOPATH

jingyuanliang:
Try a different path spec. I don't know why it worked on my dev env but not on Louhi.

/assign @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #140, opened 2022-07-22
Incorporate buildx-setup from GoogleCloudPlatform/netd@cf82473

jingyuanliang:


MrHohn (approved):
Thanks for the fix? Is this something we can upstream as well?
/lgtm
/approve

jingyuanliang:
The upstream doesn't need this fix because golang natively cross-compiles. It's needed here because of the RUN entries in Dockerfile.

MrHohn:
Got it, thanks for the clarification.

PR #144, opened 2022-08-08
Remove iproute2 dependency (using `route` from net-tools instead)

jingyuanliang:
/assign @MrHohn

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -191,8 +191,18 @@ else
     sed -e "s#@ipv6SubnetOptional##g; s#@ipv6RouteOptional##g")
 fi
 
+# Format of `route` output:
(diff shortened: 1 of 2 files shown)


scripts/install-cni.sh
```diff
@@ -192,7 +192,7 @@ else
 fi
 
 # Set mtu
-default_nic=$(ip route get 8.8.8.8 | sed -nr "s/.*dev ([^\ ]+).*/\1/p")
+default_nic=$(route -n | grep -E '^0\.0\.0\.0\s+\S+\s+0\.0\.0\.0' | grep -oE '\S+$')
```
MrHohn, line 195: For easier future reference, could you put a quick comment on what this is matching? This was not immediately clear to me until I tried out the command and learned the output format.

MrHohn (approved):
Thanks!
/lgtm
/approve

PR #145, opened 2022-08-08
Replace call to `expr` with bash-builtin operation

jingyuanliang:
/assign @MrHohn


MrHohn (approved):
Thanks!
/lgtm
/approve

PR #146, opened 2022-08-09
Move netd-init to distroless-iptables with manually added binaries

jingyuanliang:
/assign @MrHohn 
/hold


jingyuanliang:
Notes:
- This should be merged after #147.
- Third party versions need to be bumped (at least toybox 0.8.8).

MrHohn (approved):
As a follow up - could you try to put more comments in the updated Dockerfile.in file as some steps are not as obvious as before?
/lgtm
/approve

PR #147, opened 2022-08-17
Setup IPV6 when node-local stack type label is present

synth-beat:
ENABLE_IPV6 is a global flag that will set the stack type for ALL nodes. If the stack type label is present, use that instead to control node-specific stack type.

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -145,7 +145,10 @@ if [ "$ENABLE_MASQUERADE" == "true" ]; then
   fi
 fi
 
-if [ "${ENABLE_PRIVATE_IPV6_ACCESS:-}" == "true" ] || [ "$ENABLE_IPV6" == "true" ]; then
+STACK_TYPE=$(printf '%s' "$response" | jq '.metadata.labels."cloud.google.com/gke-stack-type"')
+echo "Node stack type label: '${STACK_TYPE:-}'"
+
+if [ "${ENABLE_PRIVATE_IPV6_ACCESS:-}" == "true" ] || [ "$ENABLE_IPV6" == "true" ] || [ "${STACK_TYPE:-}" == '"IPV4_IPV6"' ]; then
   node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq -r '.ipv6s[0]' ) ||:
 
   if [ -n "${node_ipv6_addr:-}" ] && [ "${node_ipv6_addr}" != "null" ]; then


jingyuanliang:
Do you have more context about the intention of this change? Changes like this have to be rolled out carefully.

synth-beat:
Marking as draft for today to discuss proper interaction with ENABLE_IPV6

scripts/install-cni.sh
```diff
@@ -145,7 +145,9 @@ if [ "$ENABLE_MASQUERADE" == "true" ]; then
   fi
 fi
 
-if [ "${ENABLE_PRIVATE_IPV6_ACCESS:-}" == "true" ] || [ "$ENABLE_IPV6" == "true" ]; then
+STACK_TYPE=$(printf '%s' "$response" | jq '.metadata.labels."cloud.google.com/gke-stack-type"')
+
+if [ "${ENABLE_PRIVATE_IPV6_ACCESS:-}" == "true" ] || [[ "${STACK_TYPE:-}" == "IPV4_IPV6" ] || [[ -z "$STACK_TYPE" && "$ENABLE_IPV6" == "true" ]]; then
```
jingyuanliang, line 150: `"${STACK_TYPE:-}" == '"IPV4_IPV6"'` note the quotes.

synth-beat: Fixed.

synth-beat:
> Marking as draft for today to discuss proper interaction with ENABLE_IPV6

Moving forward with OR'ing the fields. ENABLE_IPV6 is needed independently of the stack type label.

jingyuanliang:
> > Marking as draft for today to discuss proper interaction with ENABLE_IPV6
> 
> Moving forward with OR'ing the fields. ENABLE_IPV6 is needed independently of the stack type label.

Update your PR title to say this. It goes into merge commit description.

scripts/install-cni.sh
```diff
@@ -145,7 +145,9 @@ if [ "$ENABLE_MASQUERADE" == "true" ]; then
   fi
 fi
 
-if [ "${ENABLE_PRIVATE_IPV6_ACCESS:-}" == "true" ] || [ "$ENABLE_IPV6" == "true" ]; then
+STACK_TYPE=$(printf '%s' "$response" | jq '.metadata.labels."cloud.google.com/gke-stack-type"')
+
```
sdmodi, line 149: Would be good to add a log stating the stack type on the node. This can help debug issues where we aren't seeing the correct stack type or if we are enabling v6 due to the stack type.

sdmodi:
LGTM

scripts/install-cni.sh
synth-beat: Added log line for stack type.

jingyuanliang:
/lgtm
/assign @MrHohn 

btw do we want the ":-" thing? I don't see a reason STACK_TYPE can end up undefined.

also the uppercase/lowercase in the file seems inconsistent. maybe we can sort them out at some point but not necessarily now.

synth-beat:
> /lgtm /assign Zihong Zheng
> 
> btw do we want the ":-" thing? I don't see a reason STACK_TYPE can end up undefined.
> 
> also the uppercase/lowercase in the file seems inconsistent. maybe we can sort them out at some point but not necessarily now.

The parameter expansion (":-") is intended to ensure the stack type is an empty string as a fallback if it is missing.

jingyuanliang:
I know what it is, but I don't see a case where it will be missing as a variable. If it won't be missing, this may cover other issues such as typos in the future.

"...where it will be missing *as a variable*" I mean, if the field is missing in JSON, STACK_TYPE will get string `null`; if jq fails, it will get an empty string, so it should be never missing, if I understand it correctly.

MrHohn (approved):

synth-beat:
> I know what it is, but I don't see a case where it will be missing as a variable. If it won't be missing, this may cover other issues such as typos in the future.
> 
> "...where it will be missing _as a variable_" I mean, if the field is missing in JSON, STACK_TYPE will get string `null`; if jq fails, it will get an empty string, so it should be never missing, if I understand it correctly.

Ah, yeah, it might not be necessary. We can remove it at a later date if needed.

PR #150, opened 2022-08-26
Use thockin/go-build-template@6ce4071 instead of buildx-setup in ned-init

jingyuanliang:
/assign @MrHohn
/hold


MrHohn (approved):
/lgtm
/approve

PR #151, opened 2022-08-29
Add .buildx-initialized as a prerequisite of container-%

jingyuanliang:
This is similar to GoogleCloudPlatform/netd@f8eb709 to make sure that
buildx is ready before running make for each platform in parallel.


MrHohn (approved):
/lgtm
/approve

PR #153, opened 2022-08-30
Various build fixes and improvements

jingyuanliang:
I'm placing hold while I'm running a test build locally. I think it should work well but just want to confirm. It's inside tmux and will take more than one hour. I'll cancel the hold once I see it completed successfully.

https://github.com/spdx/license-list-data is not switched at this moment because it's not available in our mirror yet. It doesn't contain executable code anyway.

/assign @MrHohn 
/hold

### DIFF
--- a/scripts/Dockerfile.in
+++ b/scripts/Dockerfile.in
@@ -17,7 +17,7 @@
 
 FROM alpine:3.16 AS curl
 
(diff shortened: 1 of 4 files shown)


MrHohn (approved):
/lgtm
/approve

PR #159, opened 2022-09-14
Run go-licenses inside container too

jingyuanliang:
It appears even running an already-built go-licenses on an old Go machine still causes issues:
```
sigs.k8s.io/json/internal/golang/encoding/json: vendor/sigs.k8s.io/json/json.go:24:2: //go:build comment without // +build comment
```
End up using two containers because they need to have different GOPATH/working directory, and I don't really want to manipulate it inside the container.

I expect that manifest-tools has less issues as it's not interacting with Go build toolchain.

Ref: https://www.reddit.com/r/golang/comments/rn25ov/encountering_gobuild_comment_without_build/

/assign @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #163, opened 2022-10-13
Replace sed commands with bash pattern substitution in install-cni.sh

jingyuanliang:
This removes all use of `sed` command from netd-init.

/assign @MrHohn

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -60,16 +60,16 @@ fi
 
 if [ -f "/host/home/kubernetes/bin/gke" ]
 then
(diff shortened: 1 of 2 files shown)


MrHohn:
@jingyuanliang question before getting these replace PRs in - what is the motivation of replacing various tools and what kind of improvement we will gain?

jingyuanliang:
All these are replaced with shell built-ins. Hoping to eliminate a few forks / external program calls. I also think the syntax is more clear in the replacements.

jingyuanliang:
Merging into https://github.com/GoogleCloudPlatform/netd/pull/169

PR #165, opened 2022-10-13
Replace route command with reading from /proc in install-cni.sh

jingyuanliang:
This removes the use of `route` command from netd-init.

/assign @MrHohn


jingyuanliang:
Merging into https://github.com/GoogleCloudPlatform/netd/pull/169

PR #167, opened 2022-10-19
Make changes to `build` GitHub workflows

jingyuanliang:
* Change `nativebuild` to run on a matrix of Go versions. This is the fastest build so we afford to run a couple more tests here. The current code is known to fail building in Go 1.17.
* Change (full) `build` to run on push only (not pull requests, i.e. just at merge) because it's too slow.

/assign @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #177, opened 2022-11-09
Separate the src_valid_mark sysctl bit from the default configs and add a flag for it.

MrHohn:
The `src_valid_mark` sysctl bit may interfere with some linux networking features (e.g. [transparent proxying](https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html)) and is not always required when enabling the hairpin traffic path (pod -> service clusterIP -> pod_itself).

This PR adds a flag to enable/disable this sysctl bit so we will have more flexibility when deploying netd.

/assign @jingyuanliang @rahulkjoshi

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -24,10 +24,11 @@ import (
 
 // NetdConfig defines the netd config
 type NetdConfig struct {
-	EnablePolicyRouting bool
-	ExcludeDNS          bool
-	EnableMasquerade    bool
-	ReconcileInterval   time.Duration
+	EnablePolicyRouting   bool
+	EnableSourceValidMark bool
+	ExcludeDNS            bool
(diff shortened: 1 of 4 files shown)


pkg/options/options.go
```diff
@@ -39,6 +40,8 @@ func NewNetdConfig() *NetdConfig {
 func (nc *NetdConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&nc.EnablePolicyRouting, "enable-policy-routing", false,
 		"Enable policy routing.")
+	fs.BoolVar(&nc.EnableSourceValidMark, "enable-source-valid-mark", true,
```
MrHohn, line 43: @jingyuanliang Per our offline discussion - updated the default to `true` to match up the previous behavior.

PR #179, opened 2022-11-23
cleanup ip6tables rules

aojea:
Some rules are no longer needed since we allow all protocols by default.

Remove a duplicate `-w` flag that is harmless but don't needed


MrHohn (approved):
Thx!
/lgtm
/approve

PR #184, opened 2022-12-13
Update containerd and cni plugins

aojea:
update containerd dependency on tools just to silent dependabot since is not really a problem

Update the rest of dependencies for a better ~higiene~  hygiene on the project


jingyuanliang:
higiene is Portuguese.

jingyuanliang:
could you upstream `tools/` changes to [thockin/go-build-template](https://github.com/thockin/go-build-template)?

MrHohn:
Well, didn't realize approve apply lgtm label as well :)

aojea:
> Well, didn't realize approve apply lgtm label as well :)

this is new , isn't it?

PR #188, opened 2023-02-10
Bump distroless-iptables to v0.2.0

jingyuanliang:
Also, switched to registry.k8s.io and added make variables to allow specifying base image source and variation at build time.

/assign @MrHohn

### DIFF
--- a/scripts/Makefile
+++ b/scripts/Makefile
@@ -30,7 +30,9 @@ ALL_PLATFORMS ?= linux/amd64 linux/arm linux/arm64 linux/ppc64le linux/s390x
 
 # The "FROM" part of the Dockerfile.  This should be a manifest-list which
 # supports all of the platforms listed in ALL_PLATFORMS.
-BASE_IMAGE ?= k8s.gcr.io/build-image/distroless-iptables:v0.1.2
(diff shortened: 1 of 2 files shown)


scripts/Makefile
```diff
@@ -30,7 +30,9 @@ ALL_PLATFORMS ?= linux/amd64 linux/arm linux/arm64 linux/ppc64le linux/s390x
 
 # The "FROM" part of the Dockerfile.  This should be a manifest-list which
 # supports all of the platforms listed in ALL_PLATFORMS.
-BASE_IMAGE ?= k8s.gcr.io/build-image/distroless-iptables:v0.1.2
+BASE_IMAGE_PREFIX ?= registry.k8s.io/build-image
```
aojea, line 33: You have REGISTRY car below that seems to be the one that override prefix

aojea: Or is only for release images?

jingyuanliang:
/hold
I'm asking them if it's to be gke-release or gke-release-staging

scripts/Makefile
aojea: Nevermind, before coffee questions 😂

PR #199, opened 2023-02-27
Change pod MTU to match cilium_wg0 MTU, if present

amol-go:
The cilium_wg0 interface is created on the GKE node if node2node encryption is enabled. This has a lower MTU as compared to eth0 inorder to accomodate encryption headers. Hence we need to reduce MTU for pod interfaces to avoid fragmentation.


MrHohn (commented):
Thanks Jingyuan and Amol for the discussion offline - LGTM retrospectively.

PR #202, opened 2023-02-28
Refactor MTU in install-cni.sh

jingyuanliang:
/assign @amol-go


MrHohn (approved):
Thanks for templating the MTU logic to help reduce future complications.
/lgtm
/approve

PR #205, opened 2023-03-13
Add subnetCount to pod_ip_metrics.go

yiningou:


yiningou:
Add ipv4 & ipv6 subnet Count and Utilization metrics to pod_ip_metrics.go
Please review my codes. Thanks!
/assign jingyuanliang
/assign MrHohn

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +361,84 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+						glog.Errorf("Error calculating total IPs for IPv4 subnet %s: %v", ipv4Subnet.String(), err)
+						return err
+					}
+					assignedIPv4 += totalIPv4
+					c.assignedIPv4AddrCount = assignedIPv4
+				}
+			}
+		}
+	}
+	if assignedIPv4 > 0 {
```
MrHohn, line 414: Discussed offline on this - let's start without this ratio first so that we are not unnecessarily consuming more CPU/memory for metrics exporting, given this is trivial to compute on the metric consumer end.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +361,84 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
 	return nil
 }
 
+func (c *podIPMetricsCollector) countTotalIPs(subnet *net.IPNet) (uint64, error) {
```
MrHohn, line 364: Chatted with both Jingyuan and Yining on this - let's explore a bit more on reusing the existing library from the OSS CNI repo.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -37,6 +39,7 @@ const (
 	ipv6
 	dual
 
+	gkeCNIConfigDir  = "/etc/cni/net.d/10-gke-ptp.conflist"
```
MrHohn, line 42: nit: this is a file instead of a directory

pkg/metrics/collector/pod_ip_metrics_test.go
```diff
@@ -269,3 +332,53 @@ func TestSetupDirectoryWatcher(t *testing.T) {
 		}
 	}
 }
+
+func TestCalculateAssignedIP(t *testing.T) {
+	// dir4 has 1 IPv4 cidr and 1 IPv6 cidr
+	dir4 := mustCreateIPAddrDir(t, "dir4")
+	mustDeleteFile(t, dir4, "lock")
+	mustDeleteFile(t, dir4, "last_allocated_ip.0")
```
MrHohn, line 340: The `lock` and `last_allocated_ip.0` files are irrelevant to the subnet IPs calculation? These seems like unneeded codes.

pkg/metrics/collector/pod_ip_metrics_test.go
```diff
@@ -269,3 +332,53 @@ func TestSetupDirectoryWatcher(t *testing.T) {
 		}
 	}
 }
+
+func TestCalculateAssignedIP(t *testing.T) {
+	// dir4 has 1 IPv4 cidr and 1 IPv6 cidr
+	dir4 := mustCreateIPAddrDir(t, "dir4")
```
MrHohn, line 338: I see now, you are trying to reuse the `mustCreateIPAddrDir()` func. Let's just create a separate helper func as they are supposedly different folders.

pkg/metrics/collector/pod_ip_metrics_test.go
```diff
@@ -32,6 +33,68 @@ func mustCreateFile(t *testing.T, dir, name, content string) {
...
+					}
+				},
+				{
+					"type": "cilium-cni"
+				}
+			]
+		}`)
+	}
+	var parsedData map[string]interface{}
+	if err := json.Unmarshal(data, &parsedData); err != nil {
```
MrHohn, line 70: It may be complicating the test logic too much here with json manipulating - could we just put the ranges into the `data` var direclty?

MrHohn (commented):
Hey Yining, thanks for the PR! The approach generally looks good to me, leaving a few comments on the details.

pkg/metrics/collector/pod_ip_metrics.go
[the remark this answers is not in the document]
jingyuanliang: This one is tricky. How will we handle the Calico case?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+	var assignedIPv4, assignedIPv6 uint64
+	networkConfigList, err := libcni.ConfListFromFile(subnetFile)
+	if err != nil {
+		glog.Errorf("Error reading network config from file %s: %v", subnetFile, err)
+		return err
+	}
+
+	var ipamConfig *allocator.IPAMConfig
+	for _, plugin := range networkConfigList.Plugins {
+		if plugin.Network.Type == "gke" {
```
jingyuanliang, line 361: Ref https://github.com/GoogleCloudPlatform/netd/blob/a9375688ce84c55389fa63a16feee07906da2400/scripts/install-cni.sh#L87-L91

Will we have / when will we have `"type": "ptp"`?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+				glog.Errorf("Error loading IPAM config: %v", err)
+				return err
+			}
+			break
+		}
+	}
+
+	for _, ranges := range ipamConfig.Ranges {
+		for _, subnets := range ranges {
+			convertedSubnet := &net.IPNet{
```
jingyuanliang, line 373: Do we have to convert this or can we just stick to `github.com/containernetworking/cni/pkg/types.IPNet`? Also consider the `rangeStart` and `rangeEnd` comment above.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
 
+func (c *podIPMetricsCollector) countTotalIPs(subnet *net.IPNet) (uint64, error) {
+	ones, bits := subnet.Mask.Size()
+	if bits == 0 {
+		return 0, fmt.Errorf("invalid IP subnet mask %v", subnet.Mask)
+	}
+	if ones == bits {
+		ones = bits - 1
+	}
+	return uint64(1 << (bits - ones)), nil
```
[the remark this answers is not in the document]
jingyuanliang: @MrHohn this is a public repo. Let's link to https://www.cni.dev/plugins/current/ipam/host-local/#network-configuration-reference directly.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
 	return nil
 }
 
+func (c *podIPMetricsCollector) countTotalIPs(subnet *net.IPNet) (uint64, error) {
```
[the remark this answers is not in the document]
jingyuanliang: Given the comment below, this should really be changed to `countIPsFromRange`.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+			if err != nil {
+				glog.Errorf("Error loading IPAM config: %v", err)
+				return err
+			}
+			break
+		}
+	}
+
+	for _, ranges := range ipamConfig.Ranges {
+		for _, subnets := range ranges {
```
jingyuanliang, line 372: The naming `subnets` is misleading.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+			convertedSubnet := &net.IPNet{
+				IP:   subnets.Subnet.IP,
+				Mask: subnets.Subnet.Mask,
+			}
+			if convertedSubnet.IP.To16() != nil && convertedSubnet.IP.To4() == nil {
+				totalIPv6, err := c.countTotalIPs(convertedSubnet)
```
jingyuanliang, line 378: These two chunks look repetitive.. I don't have a good suggestion now though. Maybe we can look back to it until we handled `rangeStart` / `rangeEnd`?

jingyuanliang (commented):
Didn't review tests because they may be changed based on how other parts change.

pkg/metrics/collector/pod_ip_metrics.go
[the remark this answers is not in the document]
aojea: check if this is useful https://github.com/kubernetes/kubernetes/blob/3cf9f66e90d560ac080687610933c712bcf37b39/pkg/registry/core/service/ipallocator/ipallocator.go#L496-L522 , I just had to do the same for kubernetes

pkg/metrics/collector/pod_ip_metrics.go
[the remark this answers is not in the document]
MrHohn: Good catch - with Calico there are a few more things to consider:
- CNI file may not be available on netd startup.
- Need to pipe through the Calico CNI name.
- Parsing is a bit different as Calico puts `{ "subnet": "usePodCidr" }` instead of a plain CIDR.

Thought about parsing the CNI template file instead but that may not work as the template file will be removed by Calico pod promptly.

It may be fine to start without supporting this case though.

pkg/metrics/collector/pod_ip_metrics.go
MrHohn: This looks like legacy logic added for the ptp->gke migration. Need to verify further though - probably on the enablement path e.g. enabling netd on an existing cluster?

pkg/metrics/collector/pod_ip_metrics.go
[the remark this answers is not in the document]
jingyuanliang: > CNI file may not be available on netd startup.

Not really. install-cni.sh waits for calico's conflist before moving ahead.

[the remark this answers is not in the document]
jingyuanliang: For reference, Calico's regular conf:
```
{
   "name":"k8s-pod-network",
   "cniVersion":"0.3.1",
   "plugins":[
      {
         "type":"calico",
         "mtu":1460,
         "log_level":"info",
         "log_file_path":"/var/log/calico/cni/cni.log",
         "datastore_type":"kubernetes",
         "nodename":"gke-calico-netd-default-pool-8ce288fc-wb07",
         "ipam":{
            "type":"host-local",
            "subnet":"usePodCidr"
         },
         "policy":{
            "type":"k8s"
         },
         "kubernetes":{
            "kubeconfig":"/etc/cni/net.d/calico-kubeconfig"
         }
      },
      {
         "type":"portmap",
         "capabilities":{
            "portMappings":true
         },
         "snat":true
      },
      {
         "type":"bandwidth",
         "capabilities":{
            "bandwidth":true
         }
      }
   ]
}
```

Direct path conf:
```
{
   "name":"gke-pod-network",
   "cniVersion":"0.3.1",
   "plugins":[
      {
         "type":"calico",
         "mtu":1460,
         "log_level":"warning",
         "log_file_path":"/var/log/calico/cni/cni.log",
         "datastore_type":"kubernetes",
         "nodename":"gke-calico-netd-v6-default-pool-fc507ad4-vmx2",
         "ipam":{
            "type":"host-local",
            "ranges":[
               [
                  {
                     "subnet":"usePodCidr"
                  }
               ],
               [
                  {
                     "subnet":"2600:2d00:4005:c4e0:a80:3a:0:0/112"
                  }
               ]
            ],
            "routes":[
               {
                  "dst":"0.0.0.0/0"
               },
               {
                  "dst":"2001:4860:8040::/42"
               }
            ]
         },
         "policy":{
            "type":"k8s"
         },
         "kubernetes":{
            "kubeconfig":"/etc/cni/net.d/calico-kubeconfig"
         }
      },
      {
         "type":"portmap",
         "capabilities":{
            "portMappings":true
         },
         "snat":true
      },
      {
         "type":"bandwidth",
         "capabilities":{
            "bandwidth":true
         }
      }
   ]
}
```

/etc/cni/net.d/10-calico.conflist

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+				}
+				assignedIPv4 += totalIPv4
+				c.assignedIPv4AddrCount = assignedIPv4
+			}
+		}
+	}
+	return nil
+}
+
+func (c *podIPMetricsCollector) updateAssignedIP(subnetFile string) error {
```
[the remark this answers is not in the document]
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+}
+
 func (c *podIPMetricsCollector) Update(ch chan<- prometheus.Metric) error {
 	if !podIPMetricsWatcherSetup {
 		if err := c.setupDirectoryWatcher(gkePodNetworkDir); err != nil {
 			glog.Errorf("setupDirectoryWatcher returned error: %v", err)
 			return nil
 		}
 	}
+	if err := c.updateAssignedIP(gkeCNIConfigDir); err != nil {
```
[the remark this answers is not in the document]
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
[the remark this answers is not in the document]
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -322,22 +337,92 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+				if err != nil {
+					glog.Errorf("Error calculating total IPs for IPv4 subnet %s: %v", convertedSubnet.String(), err)
+					return err
+				}
+				assignedIPv4 += totalIPv4
+				c.assignedIPv4AddrCount = assignedIPv4
+			}
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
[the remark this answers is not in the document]
yiningou: Done

pkg/metrics/collector/pod_ip_metrics_test.go
yiningou: Done

pkg/metrics/collector/pod_ip_metrics_test.go
```diff
@@ -269,3 +332,53 @@ func TestSetupDirectoryWatcher(t *testing.T) {
...
+			wantAssignedIPv4Count: 256,
+			wantAssignedIPv6Count: 0,
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			mc := podIPMetricsCollector{}
+			subnetFilePath := tc.dir + "/" + "subnet"
+			var err error
+			if err = mc.calculateAssignedIP(subnetFilePath); err != nil {
+				t.Fatalf("Got error %v while calculate assigned IP count", err)
```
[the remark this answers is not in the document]
yiningou: Done

pkg/metrics/collector/pod_ip_metrics_test.go
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
yiningou: I extract the total IP calculation outside the loop, so there will be only one calculation chunk. I hope this will simplify the code.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +333,108 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+
+	for _, ranges := range ipamConfig.Ranges {
+		for _, ipRange := range ranges {
+			convertedSubnet := &net.IPNet{
+				IP:   ipRange.Subnet.IP,
+				Mask: ipRange.Subnet.Mask,
+			}
+			totalIP, err = c.countIPsFromRange(convertedSubnet)
+			if err != nil {
+				glog.Errorf("Error calculating total IPs for IPv6 subnet %s: %v", convertedSubnet.String(), err)
```
jingyuanliang, line 402: Not just IPv6 now.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +333,108 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+				return err
+			}
+			if convertedSubnet.IP.To16() != nil && convertedSubnet.IP.To4() == nil {
+				assignedIPv6 += totalIP
+				c.assignedIPv6AddrCount = assignedIPv6
```
jingyuanliang, line 407: Maybe we can pull this out of all the loops?

pkg/metrics/collector/pod_ip_metrics.go
yiningou: I just finished testing the enable netd on an existing cluster. I found that although we can see netd enabled, the file name is different regardless of whether it is enabled before recreating a node. Therefore, I think the path should not change before recreating and always be 'ptp'.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Sure. Let me also import the package `types` as well.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -37,6 +40,7 @@ const (
 	ipv6
 	dual
 
+	gkeCNIConfigFile = "/host/etc/cni/net.d/${CNI_SPEC_NAME}"
```
MrHohn, line 43: Humm, would this work? Shouldn't we need something like `os.Getenv("CNI_SPEC_NAME")`?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -286,7 +302,7 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
 	go func() {
 		defer func() {
 			watcher.Close()
-			podIPMetricsWatcherSetup = false
+			podIPMetricsWatcherIsIntialized = false
```
MrHohn, line 305: typo - Initialized?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,104 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
 }
 
 func (c *podIPMetricsCollector) Update(ch chan<- prometheus.Metric) error {
-	if !podIPMetricsWatcherSetup {
+	if !podIPMetricsWatcherIsIntialized {
 		if err := c.setupDirectoryWatcher(gkePodNetworkDir); err != nil {
 			glog.Errorf("setupDirectoryWatcher returned error: %v", err)
 			return nil
 		}
+		//calculate assigned IP count
```
MrHohn, line 420: Doesn't look like this comment provides any value :)

pkg/metrics/collector/pod_ip_metrics_test.go
```diff
@@ -269,3 +338,46 @@ func TestSetupDirectoryWatcher(t *testing.T) {
...
+			wantAssignedIPv4Count: 254,
+			wantAssignedIPv6Count: 0,
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			mc := podIPMetricsCollector{}
+			subnetFilePath := tc.dir
+			var err error
+			if err = mc.calculateAssignedIP(subnetFilePath); err != nil {
+				t.Fatalf("Got error while calculate assigned IP count: %v", err)
```
MrHohn, line 373: nit: while calculating

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,104 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
 	if err != nil {
 		glog.Errorf("Failed to add watcher for directory %s: %v", dir, err)
 	}
-	podIPMetricsWatcherSetup = true
+	podIPMetricsWatcherIsIntialized = true
+	return nil
+}
+
+func (c *podIPMetricsCollector) countIPsFromRange(subnet types.IPNet) (uint64, error) {
```
MrHohn, line 341: The comment from https://github.com/kubernetes/kubernetes/blob/3cf9f66e90d560ac080687610933c712bcf37b39/pkg/registry/core/service/ipallocator/ipallocator.go#L496-L522 is informative as it makes the code more readable - would be great to keep it as well.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,104 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+		c.assignedIPv6AddrCount += totalIP
+	} else if subnet.IP.To4() != nil {
+		c.assignedIPv4AddrCount += totalIP
+	}
+}
+
+func (c *podIPMetricsCollector) calculateAssignedIP(subnetFile string) error {
+	networkConfigList, err := libcni.ConfListFromFile(subnetFile)
+	if err != nil {
+		glog.Errorf("Error reading network config from file %s: %v", subnetFile, err)
```
MrHohn, line 377: It looks are we are double logging - the error would be logged from the caller of `calculateAssignedIP` already so we could simply return error here.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,104 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+
+	var ipamConfig *allocator.IPAMConfig
+	gkePluginFound := false
+	for _, plugin := range networkConfigList.Plugins {
+		if plugin.Network.Type == "gke" {
+			gkePluginFound = true
+			ipamConfig, _, err = allocator.LoadIPAMConfig(plugin.Bytes, "")
+			if err != nil {
+				glog.Errorf("Error loading IPAM config: %v", err)
+				return err
```
MrHohn, line 389: ditto - let's return the error directly.
```
return fmt.Errorf("error loading IPAM config: %v", err)
```

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,104 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+	if !gkePluginFound {
+		return fmt.Errorf("gke plugin not found in network config")
+	}
+
+	for _, ranges := range ipamConfig.Ranges {
+		for _, ipRange := range ranges {
+			totalIP, err := c.countIPsFromRange(ipRange.Subnet)
+			if err != nil {
+				glog.Errorf("Error calculating total IPs for subnet %s: %v", ipRange.Subnet.IP.String(), err)
+				return err
```
MrHohn, line 406: ditto:
```
return fmt.Errorf("error calculating total IPs for subnet %s: %v", ipRange.Subnet.IP.String(), err)
```

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,104 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
 	ch <- prometheus.MustNewConstMetric(dualStackCountDesc, prometheus.GaugeValue, float64(c.dualStackCount))
 	ch <- prometheus.MustNewConstMetric(dualStackErrorCountDesc, prometheus.GaugeValue, float64(c.dualStackErrorCount))
 	ch <- prometheus.MustNewConstMetric(duplicateIPCountDesc, prometheus.GaugeValue, float64(c.duplicateIPCount))
-
 	ch <- prometheus.MustNewConstMetric(ipReuseMinDesc, prometheus.GaugeValue, float64(c.reuseIPs.min))
 	ch <- prometheus.MustNewConstMetric(ipReuseAvgDesc, prometheus.GaugeValue, c.reuseIPs.sum/float64(c.reuseIPs.count))
 	ch <- prometheus.MustNewConstHistogram(ipReuseHistogramDesc, c.reuseIPs.count, c.reuseIPs.sum, c.reuseIPs.buckets)
-
+	ch <- prometheus.MustNewConstMetric(assignedIPv4AddrCountDesc, prometheus.GaugeValue, float64(c.assignedIPv4AddrCount))
+	ch <- prometheus.MustNewConstMetric(assignedIPv6AddrCountDesc, prometheus.GaugeValue, float64(c.assignedIPv6AddrCount))
```
MrHohn, line 435: Can we only export these two metrics when `calculateAssignedIP()` succeeded? so we are not exposing 0 value when it failed e.g. the calico case.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done. I have cc'd my experiment process to you all. Thank you for pointing this out.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done.

pkg/metrics/collector/pod_ip_metrics_test.go
yiningou: Done.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done.

pkg/metrics/collector/pod_ip_metrics.go
yiningou: Done.

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,116 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+	if !podIPMetricsWatcherIsInitialized {
 		if err := c.setupDirectoryWatcher(gkePodNetworkDir); err != nil {
 			glog.Errorf("setupDirectoryWatcher returned error: %v", err)
 			return nil
 		}
+		cniSpecName := os.Getenv("CNI_SPEC_NAME")
+		if cniSpecName == "" {
+			glog.Errorf("CNI_SPEC_NAME environment variable is not set")
+		} else {
+			subnetFile := "/host/etc/cni/net.d/" + cniSpecName
```
MrHohn, line 427: Let's make `"/host/etc/cni/net.d/"` a constant.

MrHohn (commented):
Thanks Yining, two minor comments. Could you fix the commit message to remove the redundant message from the fix-ups?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -91,18 +94,31 @@ var (
 		"Indicates the IP reuse duration in millisecond for all IPs.",
 		nil, nil,
 	)
-	podIPMetricsWatcherSetup = false
+	assignedIPv4AddrCountDesc = prometheus.NewDesc(
+		"IPv4_assigned_count",
```
jingyuanliang, line 98: maybe keep this all lowercase for consistency

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,116 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+	podIPMetricsWatcherIsInitialized = true
+	return nil
+}
+
+// hostsPerNetwork returns the number of available hosts in a subnet.
+// The max number is limited by the size of an uint64.
+// Number of hosts is calculated with the formula:
+// IPv4: 2^x – 2, not consider network and broadcast address
+// IPv6: 2^x - 1, not consider network address
+// where x is the number of host bits in the subnet.
```
jingyuanliang, line 346: do we consider full RangeStart/End support, instead of just the default values?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,116 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+	if bits <= ones {
+		return 0, fmt.Errorf("invalid subnet mask: %v", subnet.Mask)
+	}
+	// this checks that we are not overflowing an int64
+	if bits-ones >= 64 {
+		return math.MaxUint64, nil
+	}
+	max := uint64(1) << uint(bits-ones)
+	// Don't use the network's ".0" address,
+	if max == 0 {
```
jingyuanliang, line 358: looks like we'll never run into this case?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,116 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+		return 0, fmt.Errorf("subnet includes only the network address")
+	}
+	max--
+	if strings.Contains(subnet.IP.String(), ".") {
```
jingyuanliang, line 362: this condition looks hacky

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,116 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
...
+			}
+			c.updateAssignedIPs(ipRange.Subnet, totalIP)
+		}
+	}
 	return nil
 }
 
 func (c *podIPMetricsCollector) Update(ch chan<- prometheus.Metric) error {
-	if !podIPMetricsWatcherSetup {
+	if !podIPMetricsWatcherIsInitialized {
```
jingyuanliang, line 418: can we extract all these into a separate function?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -91,18 +94,31 @@ var (
...
+		"Indicates the total Ipv4 IPs assigned to the subnetwork.",
+		nil, nil,
+	)
+	assignedIPv6AddrCountDesc = prometheus.NewDesc(
+		"IPv6_assigned_count",
+		"Indicates the total Ipv6 IPs assigned to the subnetwork.",
+		nil, nil,
+	)
+	podIPMetricsWatcherIsInitialized = false
+	assignedIPExported               = false
```
jingyuanliang, line 108: we probably want to put these inside struct podIPMetricsCollector.

jingyuanliang:
I'm now thinking, why are we doing it by parsing the CNI config file from the beginning?

I previously thought, by doing this we can probably handle more different scenarios where the CNI is configured by some different party or in a different layout, but in the current implementation, what it can handle is pretty limited to the CNI config written by install-cni.sh using the default cni_spec_template (which is even not a part of this repo), and it currently just populates with `.spec.podCIDR`.

Given this, what about repeating what install-cni.sh does here and read `.spec.podCIDR` here then export the range size? This will happen to work well with calico's config too. By the way, is the control plane currently populating `podCIDRs` correctly for IPv6 and what about directpath?

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -318,26 +334,116 @@ func (c *podIPMetricsCollector) setupDirectoryWatcher(dir string) error {
 	if err != nil {
 		glog.Errorf("Failed to add watcher for directory %s: %v", dir, err)
 	}
-	podIPMetricsWatcherSetup = true
+	podIPMetricsWatcherIsInitialized = true
+	return nil
+}
+
+// hostsPerNetwork returns the number of available hosts in a subnet.
```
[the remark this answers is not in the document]
yiningou: > I'm now thinking, why are we doing it by parsing the CNI config file from the beginning?
> 
> I previously thought, by doing this we can probably handle more different scenarios where the CNI is configured by some different party or in a different layout, but in the current implementation, what it can handle is pretty limited to the CNI config written by install-cni.sh using the default cni_spec_template (which is even not a part of this repo), and it currently just populates with `.spec.podCIDR`.
> 
> Given this, what about repeating what install-cni.sh does here and read `.spec.podCIDR` here then export the range size? This will happen to work well with calico's config too. By the way, is the control plane currently populating `podCIDRs` correctly for IPv6 and what about directpath?

I apologize for not addressing your comment earlier, as I was focusing on fixing the above-mentioned comments. I'd be more than happy to discuss this issue further and explore possible solutions together. How about we schedule a quick meeting or have a discussion session?

MrHohn:
Yining, Jingyuan and myself did a quick sync on this and agreed going with `.spec.podCIDR` and `.spec.podCIDRs` seems like a cleaner and more consistent way to go with less dependencies to be added.

Regarding directpath - since it doesn't have an angle into IP allocation it should be fine to put it aside (it won't show up in `podCIDR`).

pkg/metrics/collector/pod_ip_metrics.go
```diff
@@ -266,6 +304,71 @@ func (c *podIPMetricsCollector) fillBuckets(diff uint64) {
...
+	var allCIDRs []string
+	if node.Spec.PodCIDR != "" {
+		allCIDRs = append(allCIDRs, node.Spec.PodCIDR)
+	}
+	for _, podCIDR := range node.Spec.PodCIDRs {
+		if len(allCIDRs) > 0 && allCIDRs[0] == podCIDR {
+			continue
+		}
+		allCIDRs = append(allCIDRs, podCIDR)
+	}
```
MrHohn, line 357: What about something like:
```
	for _, podCIDR := range node.Spec.PodCIDRs {
		allCIDRs = append(allCIDRs, podCIDR)
	}
        if len(allCIDRs) == 0 {
                 if node.Spec.PodCIDR != "" {
		        allCIDRs = append(allCIDRs, node.Spec.PodCIDR)
                 }
        }
```

MrHohn (approved):
Thanks Yining! One minor comment and LGTM.
/approve

MrHohn (approved):
Will keep the hold in case Jingyuan wants to take a final look.
/lgtm

PR #208, opened 2023-04-08
Send "expected" errors from cilium_ready in install-cni.sh to stdout

jingyuanliang:
Don't let them go to stderr and show up in logging as errors causing unexpected and unrelated support requests sent to us.

/assign @MrHohn


MrHohn (approved):
Thanks!
/lgtm
/approve

PR #211, opened 2023-04-27
Expose more conntrack-related error as metrics

yiningou:


yiningou:
Expose more conntrack-related error as metrics.
Please review my codes. Thanks!
/assign jingyuanliang
/assign MrHohn

jingyuanliang:
I'm thinking if we should still permute all metrics like this as the list grows, or switch to a map potentially with an allowlist to capture all metrics. @MrHohn thoughts?

MrHohn:
> I'm thinking if we should still permute all metrics like this as the list grows, or switch to a map potentially with an allowlist to capture all metrics. @MrHohn thoughts?

In this case I won't expect the list to grow in a significant way - these are the conntrack stats that exist for years. So I'm fine with anything simple here.

pkg/metrics/collector/conntrack.go
```diff
@@ -81,13 +97,26 @@ func parseHeader(line string) (*conntrackIndices, error) {
...
 		case "drop":
 			indices.drop = i
+		case "early_drop":
+			indices.earlyDrop = i
+		case "search_researt":
+			indices.searchRestart = i
 		}
 	}
-	if indices.insertFailed == -1 || indices.drop == -1 {
+	if indices.found == -1 || indices.invalid == -1 ||
```
MrHohn, line 116: Looks like this won't work as we didn't explicitly assign the other fields `-1` initially?

jingyuanliang:
/lgtm
wait for @MrHohn to approve.

pkg/metrics/collector/conntrack.go
yiningou: Done. Set all the indices to `-1` before this judgment.

MrHohn (approved):
Thanks!
/lgtm

PR #219, opened 2023-07-11
Enable route MTU in cilium cni

sugangli:
This PR turns on the route MTU feature (https://github.com/cilium/cilium/pull/26495) when cilium-cni is in chaining mode. `route MTU` = `link MTU` - `tunnel overhead`. If cilium agent is not aware of any tunneling, then tunnel overhead will be zero.

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -113,7 +113,7 @@ fi
 
 if [ "${ENABLE_CILIUM_PLUGIN}" == "true" ]; then
   echo "Adding Cilium plug-in to the CNI config."
-  cni_spec=${cni_spec//@cniCiliumPlugin/, {\"type\": \"cilium-cni\"\}}
+  cni_spec=${cni_spec//@cniCiliumPlugin/, {\"type\": \"cilium-cni\", \"enable-route-mtu\": true\}}
   # inotify calls back to the beginning of this script.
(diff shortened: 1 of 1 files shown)


jingyuanliang:
You don't have to test it now, but if you know it - considering you made https://github.com/cilium/cilium/pull/26495 - how will it interact with other tunnels such as egress gateway?

Also will something break if we have a newer netd (with this change) and an older anetd (without https://github.com/cilium/cilium/pull/26495)?

sugangli:
> You don't have to test it now, but if you know it - considering you made [cilium/cilium#26495](https://github.com/cilium/cilium/pull/26495) - how will it interact with other tunnels such as egress gateway?
> 

I am assuming it relies on normal routing instead of packet encapsulation? Setting route MTU is the standard way of cilium CNI, it should cover all the cases with pkt header overhead within cilium. 

> Also will something break if we have a newer netd (with this change) and an older anetd (without [cilium/cilium#26495](https://github.com/cilium/cilium/pull/26495))?
>

Yes, is there a way we could "version trait" this?

aojea:
> Yes, is there a way we could "version trait" this?

can we do the exercise of starting an old version setting this new config? 
it looks to me that the json parsing should omit the unknown field and cilium work normally, no?

sugangli:
Good point. Just tested it. Everything works normally, no route MTU is set, but no parsing error emits.

PR #220, opened 2023-08-03
Use curl-healthcheck for cilium to ensure its readiness before writin…

Ramkumar-K:
…g cni config.

The netd-init container currently writes the cni and this signals node-readiness to the kubelet.
However, the current check is only for the presence of the cilium-cni file, which is not a guarantee of cilium pod's readiness.
From the code, it seems that if the healthz returns error-code 200, then this indicates that all cilium subsystems are ok, which is a better signal for node-readiness.

This was tested by uploading a private a image and testing it in a sandbox cluster by resizing the number of nodes in a cluster and verifying the logs.

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -116,6 +116,17 @@ if [ "${ENABLE_CILIUM_PLUGIN}" == "true" ]; then
   cni_spec=${cni_spec//@cniCiliumPlugin/, {\"type\": \"cilium-cni\", \"enable-route-mtu\": true\}}
   # inotify calls back to the beginning of this script.
   inotify /host/home/kubernetes/bin cilium-cni "$0" cilium_ready
+
+  HEALTHZ_PORT="${CILIUM_HEALTHZ_PORT:-9879}"
+  # Wait upto 60s for the cilium pod to report healthy.
+  if curl -fsSm 1 --retry 60 --retry-all-errors --retry-max-time 60 --retry-delay 1 \
+      -o /dev/null --stderr - \
+      http://localhost:"${HEALTHZ_PORT}"/healthz; then
+    echo "Cilium healthz reported success."
+  else
+    echo "Cilium not yet ready. Continuing anyway."
+  fi
+
   echo "Cilium plug-in now confirmed as ready."
 else
   echo "Not using Cilium plug-in."


Ramkumar-K:
@aojea @jingyuanliang  tagging you to review this PR based on our discussion in the document

scripts/install-cni.sh
```diff
@@ -30,6 +30,23 @@ cilium_ready() {
   # Send errors to stdout since they're "expected" errors.
   # This redirection doesn't affect exit status after execution.
   CNI_COMMAND=VERSION /host/home/kubernetes/bin/cilium-cni 2>&1
+
+  HEALTHZ_PORT="${CILIUM_HEALTHZ_PORT:-9879}"
+  # Wait upto 60s for the cilium pod to report healthy.
+  for ((i=0;i<=60;++i)); do
+    echo "Checking cilium health iteration ${i}"
+    curl -w %{http_code} localhost:${HEALTHZ_PORT}/healthz | grep 200
```
aojea, line 38: better to use the complete url as `http://localhost`

```
curl -w %{http_code} http://localhost:${HEALTHZ_PORT}/healthz | grep 200

200
```

jingyuanliang:
I would structure the code in another way, but my question about "can we make cilium (seen as a whole component) write the CNI binary only when it's ready" seems not answered.

jingyuanliang:
As for restructuring, I would do something like:
`timeout 60s bash -c 'while ! curl -fsSm 1 -o /dev/null http://localhost:${HEALTHZ_PORT}/healthz; do sleep 1s; done'`
after the inotify call, if we're going with this approach.

jingyuanliang:
I might also move this chunk to the end of the script, so other parts can work before cilium is ready.

aojea:
> I might also move this chunk to the end of the script, so other parts can work before cilium is ready.

I'm not familiar with the script, but bear in mind that we can not shell out the cni config file until cilium is ready

jingyuanliang:
> > I might also move this chunk to the end of the script, so other parts can work before cilium is ready.
> 
> I'm not familiar with the script, but bear in mind that we can not shell out the cni config file until cilium is ready

It can be just before the final writeout.

jingyuanliang:
Found https://github.com/cilium/cilium/commit/d45999d5bded78590c19a670601f9c486dc0cba0 https://github.com/cilium/cilium/commit/bf060c972f8d63eb7c525fc90860928979ae962c

jingyuanliang:
This reminds me of one thing: is write-cni-conf-when-ready available in our cilium?

scripts/install-cni.sh
```diff
@@ -30,6 +30,20 @@ cilium_ready() {
   # Send errors to stdout since they're "expected" errors.
   # This redirection doesn't affect exit status after execution.
   CNI_COMMAND=VERSION /host/home/kubernetes/bin/cilium-cni 2>&1
+
+  # Wait upto 60s for the cilium pod to report healthy.
+  echo "Waiting for upto 60s for cilium to report readiness"
```
jingyuanliang, line 35: Move this after the inotify call instead of inside it. We don't want to start checking this before cilium-cni binary is written, I think.

scripts/install-cni.sh
```diff
@@ -30,6 +30,20 @@ cilium_ready() {
...
+  timeout 60s \
+    bash -c \
```
jingyuanliang, line 37: Using a heredoc is probably better than dealing with all the escapes.

scripts/install-cni.sh
```diff
@@ -30,6 +30,20 @@ cilium_ready() {
...
+      'HEALTHZ_PORT="${CILIUM_HEALTHZ_PORT:-9879}" ; \
+      echo "Querying cilium healthz on port ${HEALTHZ_PORT}" ; \
+      while ! curl -fsSm 1 -o /dev/null http://localhost:${HEALTHZ_PORT}/healthz; do \
```
jingyuanliang, line 40: Let's move the stderr output to stdout, so it doesn't go to "ERRORS" in logs confusing users. Either do it here or on the whole timeout call.

scripts/install-cni.sh
```diff
@@ -30,6 +30,20 @@ cilium_ready() {
...
+      echo "Cilium not yet ready"; \
+      sleep 1; \
+      done'
```
jingyuanliang, line 43: Maybe add an output when we find cilium ready before the timeout.

jingyuanliang:
Comments apply only if we decide to go with this approach. Also please rebase or make the commit history cleaner somehow.

aojea:
> @aojea Curious how Cilium is handling this currently on upstream - it reads like their latest script only installs the binary: https://github.com/cilium/cilium/blob/0b1e3f765ba64cfdf2058ad5081b640f3f779287/plugins/cilium-cni/install-plugin.sh

I think they just "cheat" and don't care about the first pod startup latency, so first pods keep retrying until the agent gets ready ... 🤷

Ramkumar-K:
Have updated the script as per Jingyuan's suggestion. Please take a look

jingyuanliang:
Played around and found this - might be even better:

```
if curl -fsSm 1 --retry 60 --retry-all-errors --retry-max-time 60 --retry-delay 1 -o /dev/null --stderr - http://localhost:${HEALTHZ_PORT}/healthz; then
    echo "Cilium is now started"
else
    echo "Cilium is not ready in time; continuing anyway"
fi
```

Explanation: treat http errors as failures to be retried; silent output except for failures and output them to stdout; retry all errors (including "seemingly" non-transient) up to 60 seconds (total wall time) or 60 requests (shouldn't hit this); for each request if it's not completed in 1 second, terminate it as a failure and wait for the next retry; delay 1 second between each retry. Print success/failure message based on whether we finally got a successful retry or retried until timeout (60s).

scripts/install-cni.sh
```diff
@@ -30,6 +30,21 @@ cilium_ready() {
   # Send errors to stdout since they're "expected" errors.
   # This redirection doesn't affect exit status after execution.
   CNI_COMMAND=VERSION /host/home/kubernetes/bin/cilium-cni 2>&1
+
```
jingyuanliang, line 33: Still, could you move this outside cilium_ready() to line 119 (before the change)? This is actually a callback of inotify check and it's called repeatedly before cilium-cni binary is written.

jingyuanliang:
nit: fix line length in the commit message so it doesn't unexpectedly wrap.

scripts/install-cni.sh
Ramkumar-K: Done.

scripts/install-cni.sh
Ramkumar-K: Done

scripts/install-cni.sh
Ramkumar-K: Done

PR #221, opened 2023-08-08
Move the wait for Cilium readiness to the end

jingyuanliang:
Do it just before writing out CNI config so other steps can complete earlier without waiting for Cilium.


MrHohn (approved):
/lgtm

PR #225, opened 2023-09-13
install-cni: Generalize cilium_ready in preparation for ASM integration

jingyuanliang:
Didn't test.

/assign @MrHohn 
/assign @aojea 
/hold

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -21,15 +21,16 @@ calico_ready() {
   compgen -G "/host/etc/cni/net.d/*calico*.conflist"
 }
 
-cilium_ready() {
-  echo "Running /host/home/kubernetes/bin/cilium-cni with CNI_COMMAND=VERSION"
-  # It's necessary to try running the it instead of just checking existence because
-  # Cilium's CNI installer doesn't do atomic write (write to temporary file then move).
+cni_ready() {
+  local -r cni_bin="$1"
+  echo "Running '/host/home/kubernetes/bin/${cni_bin}' with CNI_COMMAND=VERSION"
+  # It's necessary to try running it instead of just checking existence because
+  # the CNI installer might not do atomic write (write to temporary file then move).
   echo "(errors are expected during bootstrap; will retry until success)"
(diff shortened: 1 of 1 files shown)


scripts/install-cni.sh
```diff
@@ -21,15 +21,16 @@ calico_ready() {
   compgen -G "/host/etc/cni/net.d/*calico*.conflist"
 }
 
-cilium_ready() {
-  echo "Running /host/home/kubernetes/bin/cilium-cni with CNI_COMMAND=VERSION"
-  # It's necessary to try running the it instead of just checking existence because
-  # Cilium's CNI installer doesn't do atomic write (write to temporary file then move).
+cni_ready() {
+  cni="$1"
+  echo "Running '/host/home/kubernetes/bin/$cni' with CNI_COMMAND=VERSION"
```
aojea, line 26: You know more than me, is not recommend to use the ${cni} format?

scripts/install-cni.sh
```diff
@@ -245,7 +246,7 @@ if [ "${ENABLE_CILIUM_PLUGIN}" == "true" ]; then
   echo "Cilium plug-in is in use. Holding CNI configurations until Cilium is ready."
 
   # inotify calls back to the beginning of this script.
-  inotify /host/home/kubernetes/bin cilium-cni "$0" cilium_ready
+  inotify /host/home/kubernetes/bin cilium-cni "$0" cni_ready cilium-cni
```
aojea, line 249: I got lost here :)

jingyuanliang: inotify uses argv+3 as callback, so the callback is `"$0" cni_ready cilium-cni`.

when calling back, `"$@"` is `cni_ready cilium-cni` so we're calling `cni_ready` with `cilium-cni` as `$1`.

aojea: if you want a future person with that level of bash knowledge maintain this script you should write that down in the comment XD

howardjohn (approved):

jingyuanliang:
Got a chance to test it.

/hold cancel

PR #232, opened 2023-10-12
More curl disables

jingyuanliang:
/hold


jingyuanliang:
didn't test, but I want to see how fast it compiles on GitHub.

PR #251, opened 2023-11-21
Add Istio CNI configuration to Netd

dfloresh:


scripts/install-cni.sh
```diff
@@ -121,6 +121,20 @@ else
...
 
+## Look for optional env var to know if plug-in should be enabled
+if [[ -n "${ISTIO_CNI_CONFIG:-}" ]]; then
+  ENABLE_ISTIO_PLUGIN=true
+fi
+
+# Add plug-in to config
+if [ "${ENABLE_ISTIO_PLUGIN}" == "true" ]; then
+  echo "Adding Istio plug-in to the CNI config."
+  cni_spec=${cni_spec//@cniIstioPlugin/, ${ISTIO_CNI_CONFIG}}
```
jingyuanliang, line 132: Add @cniIstioPlugin to netd-config in (the sample) netd.yaml.

scripts/install-cni.sh
```diff
@@ -121,6 +121,20 @@ else
   cni_spec=${cni_spec//@cniCiliumPlugin/}
 fi
 
+## Look for optional env var to know if plug-in should be enabled
+if [[ -n "${ISTIO_CNI_CONFIG:-}" ]]; then
+  ENABLE_ISTIO_PLUGIN=true
```
jingyuanliang, line 126: I'm not sure if we'd better have this, or just use `-n "${ISTIO_CNI_CONFIG:-}"` whenever we want to check `"${ENABLE_ISTIO_PLUGIN}" == "true"`.

scripts/install-cni.sh
dfloresh: Added @cniIstioPlugin.

howardjohn (commented):
LGTM

PR #263, opened 2024-01-09
Testing framework for install-cni template processing

marqc:
Adds tests to verify cni_spec generation by install-cni.sh script

### DIFF
--- a/scripts/install-cni-testing.md
+++ b/scripts/install-cni-testing.md
@@ -1,0 +1,82 @@
+# Testing `install-cni.sh`
+
+Testing framework for generating the CNI spec file. 
+
+## Executing tests - TL;DR
+
+```bash
+make -C scripts test
+```
+
+Output:
+
+```
+Testing install-cni.sh
+Running test [testcase/testcase-basic.sh]: PASS
+Running test [testcase/testcase-directpath.sh]: PASS
+Running test [testcase/testcase-dualstack.sh]: PASS
+Test execution log available in test.log
+```
+
+## Concept 
+
+Tests execute the `install-cni.sh` script with given set of env values (test
+config) and verifies if script runs successfully (exit code == 0) and the
+generated CNI spec file.
+
+### Test cases
+
+Each test case is defined in a separate file in `testcase/testcase-<NAME>.sh`.
+For writing a new testcase you can copy the template file
+[testcase/tpl-testcase.sh](testcase/tpl-testcase.sh).
+
+Each testcase need to implement the `verify()` function that returns `0` on
+success or non-zero otherwise.
+
+### Static test configs
+Static test configs should be stored in the `testdata/` folder.
+
+### External commands call dependencies
+
+The script mocks all calls to external commands that are used to gather data or
+execute script flow (ie. curl, route, iptables, inotify). The default set of
+mocks returns code 0 and no output as defined in the
+`init_default_syscall_mocks()` function of 
+[test-install-cni.sh](test-install-cni.sh).
+
+For each test case the relevant set of mocks should be configured in the
+`before_test()` function. If given system function is called in multiple
+contexts, you should verify the arguments list and adjust the response
+accordingly.
+
+### Filesystem dependencies
+
+The `install-cni.sh` script assumes some paths and files to exist within
+the filesystem.
+
+The following resources are mounted for test execution to meet these requirements:
+
+* `test.out` folder is mounted as `/host/etc/cni/net.d` and will contain all
+  generated CNI spec files. The file name will match the testcase name.
+* `testdata/token` is mounted as `/var/run/secrets/kubernetes.io/serviceaccount/token`
+  and simulates service account token
+
+**Note** There are other filesystem dependencies that are not supported yet, ie.
+`/sys/class/net/$nic/mtu`.
+
+### Test execution
+
+Tests are executed in the same docker image that is used for a target script
+execution. This ensures the environment (shell, commands) will be the same as
+during the `install-cni.sh` script execution in target environment.
+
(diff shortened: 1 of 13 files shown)


aojea:
This is impressive, let´s get a run on CI

aojea:
@jingyuanliang why are the githb actions not running?

jingyuanliang:
Add those new shell scripts to the target of shellcheck?

aojea:
> Add those new shell scripts to the target of shellcheck?

those are tests, should we shelcheck the tests?

you are the experts here, it looks good to me but I´m very lazy with bash so you should review @jingyuanliang

jingyuanliang:
> those are tests, should we shelcheck the tests?

Hmm I just had some nitpicks and wanted shellcheck to say it for me.

jingyuanliang:
Okay I just tried it. shellcheck gives many useless findings. Don't bother to add it then.

marqc:
I decided to add it to shellcheck. Even if findings are not important or false-positive, having the check in pipeline can prevent us from introducing bugs later on.

I also want to add a readme file with documentation on the testing framework (how it works, how to use it). I will update the PR soon.

marqc:
@jingyuanliang please take a look, it's ready for review now

scripts/testcase/tpl-testcase.sh
```diff
@@ -0,0 +1,33 @@
...
+export ENABLE_BANDWIDTH_PLUGIN=false
+export ENABLE_CILIUM_PLUGIN=false
+export ENABLE_MASQUERADE=false
+export ENABLE_IPV6=false
+
+export CNI_SPEC_TEMPLATE='{}'
+
+# Method to be invoked before install-cni.sh call
+# You can configure required mocks here
+function before_test() {
```
jingyuanliang, line 18: It looks like the pattern of url matching in `curl` and `route` mocking is everywhere. Add them to the template?

scripts/testdata/expected-directpath.json
```diff
@@ -0,0 +1,25 @@
...
+    {
+      "type": "ptp",
+      "mtu": 1460,
+      "ipam": {
+        "type": "host-local",
+        "ranges": [
+          [{"subnet": "10.52.1.0/24"}], [{"subnet": "2600:1900:4000:318:0:7:0:0/112"}]
+        ],
+        "routes": [
+          {"dst": "0.0.0.0/0"}, {"dst": "::/0"}
```
jingyuanliang, line 14: This is not the expected directpath (meaning you didn't get the env right).

Directpath doesn't have a default IPv6 route. (only one to directpath prefix)

jingyuanliang:
I can think of several other test cases, but we can start with the current ones.

scripts/test-install-cni.sh
```diff
@@ -0,0 +1,97 @@
...
+  # shellcheck disable=SC1090 source-path=testcase
+  . ./"$testcase"
+
+  # initializing testcase mocks
+  before_test
+
+  # running install-cni script
+  /bin/bash ./install-cni.sh >>test.log 2>>test.log
+  exit_code="$?"
+  if [ "0" != "${exit_code}" ] ; then #>/dev/null 2>&1
```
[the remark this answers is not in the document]
marqc: leftover from debugging, removed

scripts/testcase/testcase-basic.sh
```diff
@@ -0,0 +1,67 @@
...
+export ENABLE_IPV6=false
+
+CNI_SPEC_TEMPLATE=$(cat testdata/spec-template.json)
+export CNI_SPEC_TEMPLATE
+
+function before_test() {
+
+  function curl() {
+    # shellcheck disable=SC2317
+    if [[ "$*" =~ http://metadata\.google\.internal/computeMetadata/v1/instance/network-interfaces/0 ]]; then
```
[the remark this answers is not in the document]
marqc: case statement does not support regexp matching used in these comparisons

scripts/testcase/tpl-testcase.sh
marqc: ok, added to template and moved route mock implementation to set of common mocks

scripts/testdata/expected-directpath.json
marqc: so it should use env `CNI_SPEC_IPV6_ROUTE` - right?

jingyuanliang: right.

scripts/testcase/testcase-basic.sh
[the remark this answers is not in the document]
jingyuanliang: Hmm but I don't see you using regexp in the patterns.

scripts/install-cni-testing.md
```diff
@@ -0,0 +1,82 @@
...
+Testing install-cni.sh
+Running test [testcase/testcase-basic.sh]: PASS
+Running test [testcase/testcase-directpath.sh]: PASS
+Running test [testcase/testcase-dualstack.sh]: PASS
+Test execution log available in test.log
+```
+
+## Concept 
+
+Tests execute the `install_cni.sh` script with given set of env values (test
```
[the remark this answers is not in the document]
jingyuanliang: oops, sorry, .sh

scripts/testcase/testcase-basic.sh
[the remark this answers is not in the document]
marqc: I was using it for partial match, but you are right, I can achieve the same results with shell pattern matching

scripts/testdata/expected-directpath.json
marqc: fixed

PR #269, opened 2024-02-07
Add support for IPv6 pod range.

marqc:
IPv6 addresses are supported when using a simplified spec template for ipam plugin:

```
{
  "name": "gke-pod-network",
  "cniVersion": "0.3.1",
  "plugins": [
    {
      "type": "@cniType",
      "mtu": @mtu,
      "ipam": {
          "type": "host-local",
          "ranges": [
            @subnets
          ],
          "routes": [
            @routes
          ]
      }
    },
    {
      "type": "portmap",
      "capabilities": {
        "portMappings": true
      }
    }@cniBandwidthPlugin@cniCiliumPlugin@cniIstioPlugin
  ]
}
```

Compatibility of new template spec with other use cases is covered in install-cni.sh tests.


scripts/install-cni.sh
```diff
@@ -131,19 +131,119 @@ else
...
+echo "Node stack type label: '${STACK_TYPE:-}'"
 
-if [[ "${ipv4_subnet:-}" =~ ^\"[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*/[0-9][0-9]*\"$ ]]; then
+node_ipv6_addr=''
+if [ "$ENABLE_IPV6" == "true" ] || [ "${STACK_TYPE:-}" == '"IPV4_IPV6"' ]; then
+  node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq -r '.ipv6s[0]' ) ||:
+fi
+
+cni_spec=$(fillSubnetsInCniSpec "$cni_spec" "$response" "$node_ipv6_addr")
+ipv4_subnet=$(jq -r '.spec.podCIDR' <<<"$response")
```
aojea, line 232: in an ipv6 only cluster this will be an ipv6 subnet

scripts/install-cni.sh
```diff
@@ -131,19 +131,119 @@ else
...
+#   Subnet
+# Returns:
+#   0 if valid, 1 on invalid.
+#######################################
+function is_ipv4_range {
+  local IPV4_VALIDATION_REGEXP='^([0-9]{1,3}\.){3}[0-9]{1,3}\/[0-9]{1,2}$'
+  local ip=$1
+
+  if [[ "${ip:-}" =~ ${IPV4_VALIDATION_REGEXP} ]] ; then
+    return 0
```
jingyuanliang, line 146: This actually.. doesn't need an `if` and `return`. Just put `[[ "${ip:-}" =~ ${IPV4_VALIDATION_REGEXP} ]]` as the last statement and its return value is returned automatically.

scripts/install-cni.sh
```diff
@@ -131,19 +131,119 @@ else
...
+    fi
+  fi
+
+  cni_spec=${cni_spec//@subnets/$SUBNETS_REPLACEMENT}
+  cni_spec=${cni_spec//@routes/$ROUTES_REPLACEMENT}
+
+  echo "${cni_spec}"
+}
+
+STACK_TYPE=$(jq '.metadata.labels."cloud.google.com/gke-stack-type"' <<<"$response")
```
jingyuanliang, line 223: If you're touching this, could you make this `-r` and tweak the error-prone `"${STACK_TYPE:-}" == '"IPV4_IPV6"'` below?

scripts/install-cni.sh
```diff
@@ -131,19 +131,119 @@ else
...
+  local node=$2
+  local node_ipv6_addr=$3
+
+  local SUBNETS_REPLACEMENT=""
+  local ROUTES_REPLACEMENT=""
+
+  for subnet in $(jq -r '.spec.podCIDRs[]' <<<"$node") ; do
+    if is_ipv4_range "$subnet" ; then
+      if [ "" != "${SUBNETS_REPLACEMENT}" ] ; then
+        SUBNETS_REPLACEMENT="${SUBNETS_REPLACEMENT}, "
```
jingyuanliang, line 189: Do we want to use a bash array instead of this first-element logic?

scripts/install-cni.sh
```diff
@@ -131,19 +131,119 @@ else
...
+      fi
+      SUBNETS_REPLACEMENT="${SUBNETS_REPLACEMENT}[{\"subnet\": \"${subnet}\"}]"
+      ROUTES_REPLACEMENT="${ROUTES_REPLACEMENT}{\"dst\": \"::/0\"}"
+    else
+      echo "[ERROR] Subnet $subnet is not a valid IP range"
+      exit 1
+    fi
+  done
+
+  # Directpath usecase
```
jingyuanliang, line 207: Are you sure this is only for directpath case today?

scripts/install-cni.sh
jingyuanliang: Can we unify this logic together with `fillSubnetsInCniSpec`?

scripts/install-cni.sh
marqc: done

scripts/install-cni.sh
marqc: done

scripts/install-cni.sh
marqc: I'm not 100% sure, but the condition looks like this is used when either `ENABLE_IPV6` or `DUAL_STACK` is in place. `DUAL_STACK` has IPv6 subnet defined in k8s node object in `.spec.podCIDRs` so dual stack doesn't need this code path. So I'm assuming it's only needed when `ENABLE_IPV6` is in place and I'm reading the `ENABLE_IPV6` flag as "directpath".

Actually I found that ENABLE_IPV6 flag is set when either directpath or dualstack are in place:
```
"EnableIPv6":                directpathEnabled || network.IsDualStackCluster(o.Cluster),
```

So for dual-stack this code path should be specifically omited to prevent duplicate subnet entries. I'll adjust testcases to cover that.

scripts/install-cni.sh
marqc: done

scripts/install-cni.sh
marqc: I extracted all logic for old IPAM replacements to function `fillSubnetsInCniSpecLegacyTemplate`. I would rather not mix the old and new template logic together.

scripts/install-cni.sh
```diff
@@ -170,56 +297,36 @@ if [ "$ENABLE_MASQUERADE" == "true" ]; then
...
 
-STACK_TYPE=$(jq '.metadata.labels."cloud.google.com/gke-stack-type"' <<<"$response")
-echo "Node stack type label: '${STACK_TYPE:-}'"
-
-if [ "$ENABLE_IPV6" == "true" ] || [ "${STACK_TYPE:-}" == '"IPV4_IPV6"' ]; then
-  node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq -r '.ipv6s[0]' ) ||:
-
-  if [ -n "${node_ipv6_addr:-}" ] && [ "${node_ipv6_addr}" != "null" ]; then
-    echo "Found nic0 IPv6 address ${node_ipv6_addr:-}. Filling IPv6 subnet and route..."
+if [ "$POPULATE_IP6TABLES" == "true" ] ; then
```
aojea, line 300: why do we want to make this conditional?

aojea (changes requested):

scripts/install-cni.sh
marqc: Explained in code comment

```
# ip6tables need to be propagated only if IPv6 is in use in directpath, dual-stack or IPv6 clusters.
# this flag is raised if at any point IPv6 subnet is configured.
POPULATE_IP6TABLES="false"
```

scripts/install-cni.sh
```diff
@@ -130,20 +130,149 @@ else
...
+    echo "Clearing IPv6 subnet and route given IPv6 access is disabled..."
+    cni_spec=${cni_spec//@ipv6SubnetOptional/}
+    cni_spec=${cni_spec//@ipv6RouteOptional/}
+  fi
+
+}
+
+function fillSubnetsInCniSpec {
+  fillSubnetsInCniSpecV2Template "$1" "$2"
+  fillSubnetsInCniSpecLegacyTemplate "$1" "$2"
```
aojea, line 263: @jingyuanliang I ran this thought through @marqc , should we have template versions instead of relying on the parsers?

jingyuanliang (commented):
Could you add a testcase where netd is used to generate calico conf file?

scripts/install-cni.sh
```diff
@@ -130,20 +130,149 @@ else
...
+      SUBNETS_REPLACEMENT+=("[{\"subnet\": \"${subnet}\"}]")
+      ROUTES_REPLACEMENT+=("{\"dst\": \"::/0\"}")
+    else
+      echo "[ERROR] Subnet $subnet is not a valid IP range"
+      exit 1
+    fi
+  done
+
+  # Directpath use case
+  if [ "$ENABLE_IPV6" == "true" ] && [ "$ipv6_subnet_configured" == "false" ] ; then
```
jingyuanliang, line 198: Maybe, we don't need this `[ "$ENABLE_IPV6" == "true" ]` guard, because `${node_ipv6_addr:-}` can be set only when ENABLE_IPV6 or IPV4_IPV6.

scripts/install-cni.sh
```diff
@@ -130,20 +130,149 @@ else
...
+  elif is_ipv6_range "${primary_subnet:-}" ; then
+    echo "Primary IPv6 pod range detected '${primary_subnet:-}'. It will only work with new spec template."
+  else
+    echo "Response from $node_url"
+    echo "$response"
+    echo "Failed to fetch/validate PodCIDR from K8s API server, primary_subnet=${primary_subnet:-}. Exiting (1)..."
+    exit 1
+  fi
+
+  if [ "$ENABLE_IPV6" == "true" ] || [ "${STACK_TYPE:-}" == "IPV4_IPV6" ]; then
```
jingyuanliang, line 241: Same as above.

scripts/install-cni.sh
```diff
@@ -130,20 +130,149 @@ else
...
+#######################################
+function fillSubnetsInCniSpecLegacyTemplate {
+  local node=$1
+  local node_ipv6_addr=$2
+
+  local primary_subnet
+  primary_subnet=$(jq -r '.spec.podCIDR' <<<"$response")
+
+  if is_ipv4_range "${primary_subnet:-}" ; then
+    echo "PodCIDR validation succeeded: ${primary_subnet:-}"
```
jingyuanliang, line 229: I'm okay with not keeping the exact old messages. Feel free to make changes if you want.

scripts/install-cni.sh
```diff
@@ -130,20 +130,149 @@ else
...
+function fillSubnetsInCniSpecV2Template {
+  local node=$1
+  local node_ipv6_addr=$2
+
+  local SUBNETS_REPLACEMENT=()
+  local ROUTES_REPLACEMENT=()
+
+  local ipv6_subnet_configured="false"
+
+  for subnet in $(jq -r '.spec.podCIDRs[]' <<<"$node") ; do
```
jingyuanliang, line 182: Can we add some more logging for easier troubleshooting?

scripts/install-cni.sh
jingyuanliang: I don't have strong opinion.

aojea: I think my main concern is process the template twice, seems very brittle and easy to have subtle mistakes, if we can just add a sentinel that allows us to differentiate the template, we can add an args parameter to the template and have our internal versioning

https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md#network-configuration

```
{
  "cniVersion": "0.4.0",
  "name": "dbnet",
  "type": "bridge",
  "bridge": "cni0",
  "args": {
    "labels" : {
        "netdVersion" : "1.0"
    }
  },
  "ipam": {
    "type": "host-local",
    "subnet": "10.1.0.0/16",
    "gateway": "10.1.0.1"
  },
  "dns": {
    "nameservers": [ "10.1.0.1" ]
  }
}
```

jingyuanliang: Well when you said that I was thinking about a new env var. A new label doesn't work well because the template itself is not valid json and can't be easily parsed.

aojea: env var sounds good

scripts/install-cni.sh
```diff
@@ -130,20 +130,149 @@ else
...
+node_ipv6_addr=''
+if [ "$ENABLE_IPV6" == "true" ] || [ "${STACK_TYPE:-}" == "IPV4_IPV6" ]; then
+  node_ipv6_addr=$(curl -s -k --fail "http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/?recursive=true" -H "Metadata-Flavor: Google" | jq -r '.ipv6s[0]' ) ||:
 fi
 
-echo "Filling IPv4 subnet ${ipv4_subnet:-}"
-cni_spec=${cni_spec//@ipv4Subnet/[{\"subnet\": ${ipv4_subnet:-}\}]}
+fillSubnetsInCniSpec "$response" "$node_ipv6_addr"
 
 if [ "$ENABLE_MASQUERADE" == "true" ]; then
```
jingyuanliang, line 277: By the way I think this chunk can be removed... @MrHohn to verify?

MrHohn: Oh I missed this - yes this is legacy logic that is no longer in used. Don't have to do it in the same PR though.

jingyuanliang: Yeah just spotted it here since this has some major overhaul of the script.

scripts/install-cni.sh
marqc: done

scripts/install-cni.sh
marqc: removed

scripts/install-cni.sh
marqc: rewritten to better reflect current logic

scripts/install-cni.sh
marqc: Added

scripts/install-cni.sh
marqc: Added handler for env `CNI_SPEC_TEMPLATE_VERSION` that fallbacks to legacy (current) template parser if not set.

scripts/install-cni.sh
```diff
@@ -130,71 +130,185 @@ else
...
+  local node_ipv6_addr=$2
+
+  local primary_subnet
+  primary_subnet=$(jq -r '.spec.podCIDR' <<<"$response")
+
+  if is_ipv4_range "${primary_subnet:-}" ; then
+    echo "PodCIDR IPv4 detected: '${primary_subnet:-}'"
+    cni_spec=${cni_spec//@ipv4Subnet/[{\"subnet\": \"${primary_subnet:-}\"\}]}
+  elif is_ipv6_range "${primary_subnet:-}" ; then
+    echo "Primary IPv6 pod range detected '${primary_subnet:-}'. It will only work with new spec template."
```
jingyuanliang, line 240: exit 1?

scripts/install-cni.sh
```diff
@@ -130,71 +130,185 @@ else
...
+# Arguments:
+#   node json from kube-apiserver
+#   node_ipv6_addr node's IPv6 address from GCE metadata server
+#######################################
+function fillSubnetsInCniSpecLegacyTemplate {
+  local node=$1
+  local node_ipv6_addr=$2
+
+  local primary_subnet
+  primary_subnet=$(jq -r '.spec.podCIDR' <<<"$response")
```
jingyuanliang, line 234: <<<"$node"

scripts/install-cni.sh
```diff
@@ -130,71 +130,185 @@ else
...
+      exit 1
+    fi
+  done
+
+  # Directpath use case
+  if [ "$ipv6_subnet_configured" == "false" ] ; then
+    # Directpath adds IPv6 subnet and route derived from host with fixed range
+    # of /112 even when it is not specified in node's .spec.podCIDRs
+    if [ -n "${node_ipv6_addr:-}" ] && [ "${node_ipv6_addr}" != "null" ]; then
+      POPULATE_IP6TABLES="true"
```
jingyuanliang, line 210: Add one more logging here?

scripts/install-cni.sh
```diff
@@ -130,71 +130,185 @@ else
...
+  primary_subnet=$(jq -r '.spec.podCIDR' <<<"$response")
+
+  if is_ipv4_range "${primary_subnet:-}" ; then
+    echo "PodCIDR IPv4 detected: '${primary_subnet:-}'"
+    cni_spec=${cni_spec//@ipv4Subnet/[{\"subnet\": \"${primary_subnet:-}\"\}]}
+  elif is_ipv6_range "${primary_subnet:-}" ; then
+    echo "Primary IPv6 pod range detected '${primary_subnet:-}'. It will only work with new spec template."
+  else
+    echo "Response from $node_url"
+    echo "$response"
```
jingyuanliang, line 243: "$node"

marqc:
@jingyuanliang Thanks for suggestions. All applied

scripts/install-cni.sh
```diff
@@ -130,71 +130,190 @@ else
...
-    fi
 
-    if ip6tables -w -L FORWARD | grep "Chain FORWARD (policy DROP)" > /dev/null; then
-      echo "Add rules to accept all forwarded TCP/UDP/ICMP/SCTP IPv6 packets"
-      ip6tables -A FORWARD -w -p tcp -j ACCEPT
-      ip6tables -A FORWARD -w -p udp -j ACCEPT
-      ip6tables -A FORWARD -w -p icmpv6 -j ACCEPT
-      ip6tables -A FORWARD -w -p sctp -j ACCEPT
-    fi
+STACK_TYPE=$(jq -r '.metadata.labels."cloud.google.com/gke-stack-type"' <<<"$response")
```
aojea, line 276: this is the controversial part, please add a comment explaining the meaning of this label.

AFAIK since last discussion , this labels represents the cluster-stack-type as per https://kubernetes.io/docs/concepts/cluster-administration/networking/#cluster-network-ipfamilies

@sdmodi @basantsa1989 @jingyuanliang so we are all in the same place.

So the logging below as "node" stack type will be confusing, I think it is better if we use "Cluster stack type label"

scripts/install-cni.sh
```diff
@@ -130,71 +130,190 @@ else
...
+        SUBNETS_REPLACEMENT+=("[{\"subnet\": \"usePodCidr\"}]")
+        ROUTES_REPLACEMENT+=("{\"dst\": \"0.0.0.0/0\"}")
+      else
+        SUBNETS_REPLACEMENT+=("[{\"subnet\": \"${subnet}\"}]")
+        ROUTES_REPLACEMENT+=("{\"dst\": \"0.0.0.0/0\"}")
+      fi
+    elif is_ipv6_range "${subnet}" ; then
+      echo "IPv6 subnet detected in .spec.podCIDRs: '${subnet:-}'"
+      POPULATE_IP6TABLES="true"
+      echo "ip6tables will be populated because IPv6 podCIDR is configured (from .spec.podCIDRs)"
```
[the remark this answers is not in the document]
jingyuanliang: Can we just fix CCM to match the expectation?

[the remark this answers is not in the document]
sdmodi: I believe we are already working on this. CCM needs this bug fix regardless.

scripts/install-cni.sh
jingyuanliang: I thought this was to resolve migration, so the stack type of each node is frozen at node creation, not subject to change upon netd restarts / node restarts?

scripts/install-cni.sh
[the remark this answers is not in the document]
sdmodi: Yes. Exactly.

scripts/install-cni.sh
```diff
@@ -130,71 +130,190 @@ else
...
-fi
+#######################################
+# Checks if given subnet is valid IPv6 range.
+# Arguments:
+#   Subnet
+# Returns:
+#   0 if valid, 1 on invalid.
+#######################################
+function is_ipv6_range {
+  local IPV6_VALIDATION_REGEXP='^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))\/[0-9]{1,3}$'
```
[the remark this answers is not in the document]
jingyuanliang: But maybe, if not validated, we need to properly escape it before putting it into json.

scripts/install-cni.sh
```diff
@@ -130,71 +130,190 @@ else
...
+
+  for subnet in $(jq -r '.spec.podCIDRs[]' <<<"$node") ; do
+    if is_ipv4_range "${subnet}" ; then
+      echo "IPv4 subnet detected in .spec.podCIDRs: '${subnet:-}'"
+      if [ "${ENABLE_CALICO_NETWORK_POLICY}" == "true" ]; then
+        # calico uses special value `usePodCidr` instead of directly providing IP range
+        SUBNETS_REPLACEMENT+=("[{\"subnet\": \"usePodCidr\"}]")
+        ROUTES_REPLACEMENT+=("{\"dst\": \"0.0.0.0/0\"}")
+      else
+        SUBNETS_REPLACEMENT+=("[{\"subnet\": \"${subnet}\"}]")
```
jingyuanliang, line 190: `SUBNETS_REPLACEMENT+=($(jq -n --arg subnet "${subnet}" '[{"subnet": $subnet}]'))` to allow proper escaping, especially if not validated above?

scripts/install-cni.sh
```diff
@@ -130,71 +130,190 @@ else
...
+# Arguments:
+#   Subnet
+# Returns:
+#   0 if valid, 1 on invalid.
+#######################################
+function is_ipv4_range {
+  local IPV4_VALIDATION_REGEXP='^([0-9]{1,3}\.){3}[0-9]{1,3}\/[0-9]{1,2}$'
+  local ip=$1
+
+  [[ "${ip:-}" =~ ${IPV4_VALIDATION_REGEXP} ]]
```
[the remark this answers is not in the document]
jingyuanliang: Not really, because all cidrs (including ipv6 ones) are checked by this first, so there'll always be errors showing up in logs causing confusion.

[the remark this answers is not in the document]
aojea: all the cidrs coming from node.spec.cidrs have been already validated

scripts/install-cni.sh
marqc: done

scripts/install-cni.sh
marqc: i changed variable name and comment to clearly indicate that this is a CLUSTER_STACK_TYPE

scripts/install-cni.sh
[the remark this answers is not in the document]
jingyuanliang: @sdmodi do you have the ETA or timeline of the fix?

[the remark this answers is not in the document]
aojea: it is ready to merge https://github.com/kubernetes/cloud-provider-gcp/pull/652 , I'm waiting for a last pass from Sudeep

scripts/install-cni.sh
```diff
@@ -128,71 +128,190 @@ else
...
+  local ROUTES_REPLACEMENT=()
+
+  local ipv6_subnet_configured="false"
+
+  for subnet in $(jq -r '.spec.podCIDRs[]' <<<"$node") ; do
+    if is_ipv4_range "${subnet}" ; then
+      echo "IPv4 subnet detected in .spec.podCIDRs: '${subnet:-}'"
+      if [ "${ENABLE_CALICO_NETWORK_POLICY}" == "true" ]; then
+        # calico uses special value `usePodCidr` instead of directly providing IP range
+        SUBNETS_REPLACEMENT+=("[{\"subnet\": \"usePodCidr\"}]")
```
jingyuanliang, line 185: All these can go with single quotes at outside to avoid escaping.

I don't bother to change all legacy code, but I think `SUBNETS_REPLACEMENT+=('[{"subnet": "usePodCidr"}]')` is more readable in new code?

marqc: Changed in whole fillSubnetsInCniSpecV2Template function to avoid escaping where possible

PR #272, opened 2024-02-13
Remove "Config MASQUERADE rule" from install-cni.sh

jingyuanliang:
They're always already configured by node startup script nowadays.

This should ease ebpf masquerade migration happening soon.

/assign @MrHohn @yiningou

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -145,31 +145,6 @@ fi
 echo "Filling IPv4 subnet ${ipv4_subnet:-}"
 cni_spec=${cni_spec//@ipv4Subnet/[{\"subnet\": ${ipv4_subnet:-}\}]}
 
-if [ "$ENABLE_MASQUERADE" == "true" ]; then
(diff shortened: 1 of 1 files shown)


aojea:
> This should ease ebpf masquerade migration happening soon.

and what happens to the cluster that don't use ebpf masquerade?

jingyuanliang:
> > This should ease ebpf masquerade migration happening soon.
> 
> and what happens to the cluster that don't use ebpf masquerade?

Whether they use ebpf masquerade or not, this is a duplication of the setup in node startup script.

MrHohn:
> > > This should ease ebpf masquerade migration happening soon.
> > 
> > 
> > and what happens to the cluster that don't use ebpf masquerade?
> 
> Whether they use ebpf masquerade or not, this is a duplication of the setup in node startup script.

Yep, this block isn't triggered in either case.

MrHohn (approved):
/lgtm

PR #277, opened 2024-02-22
Pass CILIUM_FAST_START_NAMESPACES env var over into cilium-cni

jingyuanliang:
/assign @MrHohn 
/assign @pravk03
/hold


MrHohn (commented):
The shell script logic seems good to me but don't have the full context on fast start namespace - will wait for Praveen to review as well.

pravk03 (approved):

PR #299, opened 2024-04-11
Fix install-cni v2 template test cases broken by #297

jingyuanliang:
/assign @marqc

### DIFF
--- a/scripts/testcase/testcase-watchdog-cilium-faststart-unhealthy-v2.sh
+++ b/scripts/testcase/testcase-watchdog-cilium-faststart-unhealthy-v2.sh
@@ -59,6 +59,13 @@ function before_test() {
   }
   export -f curl
 
+  # shellcheck disable=SC2317
+  function sleep() {
+    echo "[MOCK called] sleep $*"
+    echo "[MOCK] this test expects a delay during fast start."
+  }
+  export -f sleep
+
 }
 
 function verify() {
--- a/scripts/testcase/testcase-watchdog-cilium-faststart-v2.sh
+++ b/scripts/testcase/testcase-watchdog-cilium-faststart-v2.sh
@@ -56,6 +56,13 @@ function before_test() {
   }
   export -f curl
 
+  # shellcheck disable=SC2317
+  function sleep() {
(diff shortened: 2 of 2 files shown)


marqc (approved):

scripts/testcase/testcase-watchdog-cilium-faststart-unhealthy-v2.sh
```diff
@@ -59,6 +59,13 @@ function before_test() {
   }
   export -f curl
 
+  # shellcheck disable=SC2317
+  function sleep() {
+    echo "[MOCK called] sleep $*"
+    echo "[MOCK] this test expects a delay during fast start."
```
aojea, line 65: how did the test passed before then?

jingyuanliang: Because https://github.com/GoogleCloudPlatform/netd/pull/269 was a merge, without rebasing to pick up #297 changes...

marqc: I suspect they just took much longer to pass, as it was actually sleeping/waiting for 60seconds

marqc: Ah. You're right. It should have failed on https://github.com/GoogleCloudPlatform/netd/blob/master/scripts/test-install-cni.sh#L48

marqc: @aojea I see now. There were 2 PRs merged and each of them was working at the time they were sent and verified.
#269 and #297

So only after they were merged together into the master, the tests started failing. So probably we should have rebase #269 before merging on top of  the master branch as this PR was waiting a few weeks before getting merged. The test logic works ok.

aojea: IIRC github actions does not rebase, so it test against the state of the repo that was at that point

PR #302, opened 2024-04-23
Correctly initialize with an IPv6 API server endpoint.

krzykwas:
If $KUBERNETES_SERVICE_HOST is an IPv6 address, wrap it with [], so a curl command to /api/v1/nodes succeeds.

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -95,7 +95,13 @@ else
 fi
 
 token=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
-node_url="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
+host=${KUBERNETES_SERVICE_HOST}
+# If host contains a colon (:), it is an IPv6 address, hence needs wrapping
+# with [..].
+if [[ "${host}" =~ : ]]; then
+  host="[$host]"
+fi
+node_url="https://$host:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
 response=$(curl -k -s -H "Authorization: Bearer $token" "$node_url")
 
 if [ "${MIGRATE_TO_DPV2:-}" == "true" ]; then
--- a/scripts/testcase/testcase-basic-v2-ipv4.sh
+++ b/scripts/testcase/testcase-basic-v2-ipv4.sh
@@ -1,0 +1,66 @@
+export KUBERNETES_SERVICE_HOST=10.0.0.1
+export KUBERNETES_SERVICE_PORT=443
+
+export ENABLE_CALICO_NETWORK_POLICY=false
+export ENABLE_BANDWIDTH_PLUGIN=false
+export ENABLE_CILIUM_PLUGIN=false
+export ENABLE_MASQUERADE=false
+export ENABLE_IPV6=false
+
+CNI_SPEC_TEMPLATE=$(cat testdata/spec-template-v2.json)
+export CNI_SPEC_TEMPLATE
+
+export CNI_SPEC_TEMPLATE_VERSION=2.0
+
+function before_test() {
+
+  function curl() {
+    # shellcheck disable=SC2317
+    case "$*" in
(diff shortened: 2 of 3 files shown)


marqc (commented):
Can you add a test case with IPv4 and IPv6 KUBERNETES_SERVICE_HOST and verification that a correct curl arguments are passed, generally a clone of this test case https://github.com/GoogleCloudPlatform/netd/blob/master/scripts/testcase/testcase-basic-v2.sh#L1
https://github.com/GoogleCloudPlatform/netd/blob/master/scripts/testcase/testcase-basic-v2.sh#L23

scripts/install-cni.sh
```diff
@@ -95,7 +95,13 @@ else
 fi
 
 token=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
-node_url="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
+host=${KUBERNETES_SERVICE_HOST}
+# If host constains a colon (:), it is an IPv6 address, hence needs wrapping
+# with [..].
+if [ "${host#*:}" != "$host" ]; then
```
jingyuanliang, line 101: can we say `[[ "${host}" =~ : ]]` instead? `[ "${host#*:}" != "$host" ]` is a bit obscured.

krzykwas: Thanks for the suggestion - done.

scripts/install-cni.sh
```diff
@@ -98,7 +98,7 @@ token=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
 host=${KUBERNETES_SERVICE_HOST}
 # If host constains a colon (:), it is an IPv6 address, hence needs wrapping
```
jingyuanliang, line 99: "constains" is a typo.

jingyuanliang: I fixed this myself from GitHub UI, but this PR needs squashing before merge.

scripts/testcase/testcase-basic-v2-ipv4.sh
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function verify() {
+  local expected
+  local actual
+
+  expected=$(jq -S . <"testdata/expected-basic.json")
+  actual=$(jq -S . <"/host/etc/cni/net.d/${CNI_SPEC_NAME}")
+
+  if [ "$expected" != "$actual" ] ; then
```
aojea, line 58: naive question, how is this test exercising this change?

marqc: If the target url doesn't match in the "curl mock definition" it will exit with error code.

marqc: exact lines that are important

https://github.com/GoogleCloudPlatform/netd/pull/302/files#diff-c0207b47b22434a1b87971b51204b818d1754b3fc2b34cba1030f7e904b2138dR23

https://github.com/GoogleCloudPlatform/netd/pull/302/files#diff-3b7d03490cb92caf2e0e2559443e5eca06100e97c7b4ce190daf83fac1d60d5fR23

scripts/install-cni.sh
krzykwas: Thanks! Squashed

jingyuanliang:
cc @yiningou - cl/624106013 needs similar thing.

jingyuanliang:
I'm not sure if ip-masq-agent will still be relevant in the current ipv6-only design (v6 folks could you advise?) but at least we shouldn't let it crash when it runs over the curl lines.

aojea:
> I'm not sure if ip-masq-agent will still be relevant in the current ipv6-only design (v6 folks could you advise?) but at least we shouldn't let it crash when it runs over the curl lines.

where is the ip-masq-agent using curl?

jingyuanliang:
> where is the ip-masq-agent using curl?

No it's not the kubernetes-sigs/ip-masq-agent but our enablement approach for cilium's ip-masq-agent.

aojea:
all golang code must use the net.JoinHostPort() helper to build urls, that do the right thing https://pkg.go.dev/net#JoinHostPort

jingyuanliang:
Good to know but it's another piece of bash script there...

PR #306, opened 2024-04-24
fixing typo to properly expand variable with multiple values

lutierigb:
in case there are two default routes the quotation marks will concatenate the interface names. without the quotes, each interface will be treated as a separate value in the for loop. see http://b/336186274#comment25


jingyuanliang:
Can we add a comment saying $default_nic may be multi-line specifying multiple interfaces? In case we forget this and add the quotes back later.

jingyuanliang:
https://github.com/GoogleCloudPlatform/netd/pull/308 for a better version.

lutierigb:
closing in favor of https://github.com/GoogleCloudPlatform/netd/pull/308

PR #308, opened 2024-04-25
Handle multiple default routes correctly

jingyuanliang:
/assign @sypakine
/assign @lutierigb
/assign @yvetteli0314

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -311,6 +311,7 @@ if [ "$POPULATE_IP6TABLES" == "true" ] ; then
 
   if [ "${ENABLE_CALICO_NETWORK_POLICY}" == "true" ]; then
     echo "Enabling IPv6 forwarding..."
+    # IPV6_FORWARDING_CONF override only to be used in tests.
     echo 1 > "${IPV6_FORWARDING_CONF:-/proc/sys/net/ipv6/conf/all/forwarding}"
   fi
 fi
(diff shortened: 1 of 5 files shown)


scripts/install-cni.sh
```diff
@@ -329,13 +329,16 @@ MTU_SOURCE="<default>"
 # last field, which is the interface name. We stick to using grep to avoid
 # introducing too many new dependencies.
 
-default_nic=$(route -n | grep -E '^0\.0\.0\.0\s+\S+\s+0\.0\.0\.0' | grep -oE '\S+$')
+# In the case where there are multiple default routes (e.g. multi-networking
+# with route manager), try each one starting from the lowest metric value.
+readarray -t default_nics < <(route -n | grep -E '^0\.0\.0\.0\s+\S+\s+0\.0\.0\.0' | sort -n -k5,5 | grep -oE '\S+$')
```
sypakine, line 334: not sure if there is consensus in the community, but I consider `awk` the column getter:

`... | awk 'NF{ print $NF }'`

scripts/install-cni.sh
```diff
@@ -329,13 +329,16 @@ MTU_SOURCE="<default>"
...
 
 # cilium_wg0 is the interface for node-to-node encryption. If it's available /
 # in use, it has a lower MTU than the default NIC (eth0) due to encryption headers.
-for nic in cilium_wg0 "$default_nic"; do
-  if [ -f "/sys/class/net/$nic/mtu" ]; then
-    MTU=$(cat "/sys/class/net/$nic/mtu")
+for nic in cilium_wg0 "${default_nics[@]}"; do
+  mtu_file=${SYS_CLASS_NET:-/sys/class/net}/$nic/mtu
```
sypakine, line 339: Not familiar with the testing restrictions, but I take it this was introduced because during tests you are not allowed to write to `/sys/class/net`?

If my understanding is correct, let's document that.

scripts/install-cni.sh
jingyuanliang: There's no awk available in the image.

scripts/shell-test.sh
```diff
@@ -59,7 +59,10 @@ run_test inotify_cmd
 inotify / '' /bin/cat /dev/null >/dev/null && pass || fail
 
 run_test default_nic_mtu
-[[ -f "/sys/class/net/$(route -n | grep -E '^0\.0\.0\.0\s+\S+\s+0\.0\.0\.0' | grep -oE '\S+$')/mtu" ]] && pass || fail
+[[ -f "/sys/class/net/$(route -n | grep -E '^0\.0\.0\.0\s+\S+\s+0\.0\.0\.0' | sort -n -k5,5 | grep -oE '\S+$')/mtu" ]] && pass || fail
```
sypakine, line 62: Are we able to avoid duplicating the business logic in the test?

scripts/install-cni.sh
sypakine: 😭

scripts/install-cni.sh
jingyuanliang: Is it ever possible to write to /sys/class/net anywhere?

scripts/shell-test.sh
jingyuanliang: Not easily. The purpose here is to make sure the file in /sys/class/net/x/mtu is "typically" available. I want to have the "x" extraction as close as possible to the actual logic.

jingyuanliang: I should add a `head` here though, or in a system with multiple default routes this test fails.

scripts/install-cni.sh
sypakine: Ha, fair enough -- I had never tried 😝

scripts/shell-test.sh
jingyuanliang: It turns out not possible to add a head, because we're restricted to whatever commands available in the image.

sypakine:
/lgtm

Want to hear from @yvetteli0314 with regards to multi-network compatibility.

PR #322, opened 2024-05-14
Handle RUN_CNI_WATCHDOG from all exit paths

jingyuanliang:


scripts/install-cni.sh
```diff
@@ -50,6 +51,24 @@ echo "Install-CNI ($0), Build: $BUILD"
 
 set -u -e
 
+# Try to decouple RUN_CNI_WATCHDOG and ENABLE_CILIUM_PLUGIN; don't assume
+# ENABLE_CILIUM_PLUGIN is set whenever RUN_CNI_WATCHDOG is set.
+# All exit paths should call exit_success instead of `exit 0`,
+# so the appropriate actions can be taken according to RUN_CNI_WATCHDOG.
+exit_success() {
```
pravk03, line 58: rename ? 

We do not exit when watchdog is enabled

jingyuanliang: Do you have a recommendation? I'm not good at doing naming...

jingyuanliang: Done

scripts/install-cni.sh
```diff
@@ -405,40 +419,30 @@ cilium_health_check() {
...
+# Cilium health check logic before watchdog and CNI STATUS API are introduced.
+cilium_wait_or_ignore() {
+  if cilium_health_check "${CILIUM_HEALTH_MAX_WAIT_TIME:-600}"; then
+    echo "Cilium healthz reported success."
+  else
+    echo "Cilium not yet ready. Continuing anyway."
   fi
+}
 
+write_and_success() {
```
pravk03, line 431: Maybe changing this to write_cni_conf and making a separate call to success is cleaner.

pravk03 (approved):

scripts/install-cni.sh
jingyuanliang: It's already merged. Let's leave it as-is.

PR #324, opened 2024-05-20
install-cni.sh tweaks for performance

jingyuanliang:
* Include `nice` command allowing setting lower nice values
* Output all log entries with timestamps for easier observation
* Make ENABLE_BANDWIDTH_PLUGIN optional preparing for removal
* Remove unnecessary `cat` usage and avoid `jq -c` solely for log prettiness

/assign @marqc

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -14,10 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+log() {
+  echo "$@"
+}
+
 # shellcheck disable=SC2317,SC2329 # when called with $1=calico_ready
 calico_ready() {
-  echo "Listing items matching /host/etc/cni/net.d/*calico*.conflist"
-  echo "(this action repeats during bootstrap until a match is found)"
+  log "Listing items matching /host/etc/cni/net.d/*calico*.conflist"
+  log "(this action repeats during bootstrap until a match is found)"
   # The command producing exit status must be the last command here.
   compgen -G "/host/etc/cni/net.d/*calico*.conflist"
 }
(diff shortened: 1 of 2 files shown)


scripts/install-cni.sh
```diff
@@ -209,14 +211,14 @@ function fillSubnetsInCniSpecV2Template {
...
+      log "IPv6 subnet detected in .spec.podCIDRs: '${subnet:-}'"
       POPULATE_IP6TABLES="true"
-      echo "ip6tables will be populated because IPv6 podCIDR is configured (from .spec.podCIDRs)"
+      log "ip6tables will be populated because IPv6 podCIDR is configured (from .spec.podCIDRs)"
       ipv6_subnet_configured="true"
       SUBNETS_REPLACEMENT+=("$(jq -nc --arg subnet "${subnet}" '[{"subnet": $subnet}]')")
       ROUTES_REPLACEMENT+=('{"dst": "::/0"}')
     else
-      echo "[ERROR] Subnet detected in .spec.podCIDRs '${subnet}' is not a valid IP range"
+      log "[ERROR] Subnet detected in .spec.podCIDRs '${subnet}' is not a valid IP range"
```
marqc, line 221: Maybe it's better to introduce log level on the log handler method instead. I.e. separate methods for log_debug, log_info and log_error. Also log_error might have built-in script termination with error code.

jingyuanliang: The current logs are pretty random... I just replaced all echo with log for the purpose of appending timestamps. Maybe more improvements later if we have capacity.

marqc: Sounds good! It's definitely an improvement even in current state.

scripts/install-cni.sh
```diff
@@ -14,21 +14,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+log() {
+  echo "[${EPOCHREALTIME}]" "$@"
```
sypakine, line 18: Users will ingest these logs via stackdriver or other logging system -- can we rely on those to provide the timestamps from the container logs?

jingyuanliang: I made them for easier observation from direct `kubectl logs` calls. Things like kubelet logs also have timestamps.

jingyuanliang: If it's a `$(date)` call I wouldn't do it, but `${EPOCHREALTIME}` being bash builtin I assume the overhead is negligible?

jingyuanliang: Also in case of ingestion delay - we're trying to optimize for sub-second delay here.

sypakine: The container logs already have timestamps:

```
$ kubectl logs example --timestamps
2024-05-21T18:44:08.265571760Z Hello, World!
```

This is the same timestamp provided by stackdriver in the `timestamp` field (and is distinct from `receivedTimestamp`).

jingyuanliang: Wow didn't know about this, but let's leave the `log` call for further extension (e.g. some messages to be written to stderr).

sypakine: SGTM

PR #329, opened 2024-05-24
Watch node object for podCIDR instead of crash + wait for restart

jingyuanliang:
This can save up to 10 seconds before Kubelet decides to restart the pod after the crash.

Also made some logging improvements.

/assign @marqc 
/assign @sypakine 
/hold

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -18,6 +18,11 @@ log() {
   echo "$@"
 }
 
+fatal() {
+  echo FATAL: "$@" >&2
+  exit 1
+}
+
 # shellcheck disable=SC2317,SC2329 # when called with $1=calico_ready
 calico_ready() {
   log "Listing items matching /host/etc/cni/net.d/*calico*.conflist"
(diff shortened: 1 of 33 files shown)


scripts/install-cni.sh
```diff
@@ -110,15 +115,20 @@ else
...
 # If host contains a colon (:), it is an IPv6 address, hence needs wrapping
 # with [..].
 if [[ "${host}" =~ : ]]; then
   host="[$host]"
 fi
-node_url="https://$host:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
-response=$(curl -k -s -H "Authorization: Bearer $token" "$node_url")
+node_url="https://$host:${KUBERNETES_SERVICE_PORT}/api/v1/watch/nodes/${HOSTNAME}"
+for _ in {1..3}; do
+  response=$(grep -m1 . <(curl -k -s -N -H "Authorization: Bearer $(</var/run/secrets/kubernetes.io/serviceaccount/token)" "$node_url" | jq --unbuffered -c '.object | select(.spec.podCIDR != null)'))
```
marqc, line 127: there is no timeout specified, please specify the timeout explicitly using the `timeoutSeconds` request param to avoid  the forever hanging call.

scripts/install-cni.sh
```diff
@@ -110,15 +115,20 @@ else
...
 host=${KUBERNETES_SERVICE_HOST}
 # If host contains a colon (:), it is an IPv6 address, hence needs wrapping
 # with [..].
 if [[ "${host}" =~ : ]]; then
   host="[$host]"
 fi
-node_url="https://$host:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
-response=$(curl -k -s -H "Authorization: Bearer $token" "$node_url")
+node_url="https://$host:${KUBERNETES_SERVICE_PORT}/api/v1/watch/nodes/${HOSTNAME}"
+for _ in {1..3}; do
```
marqc: nit: 1..3 is a magic number. make it a named const variable that says this is the number of retries.

scripts/install-cni.sh
```diff
@@ -110,15 +115,23 @@ else
...
+for attempt in {1..3}; do
+  log "Watching attempt #${attempt} at ${node_url}"
+  response=$(grep -m1 . <(curl -k -s -N -H "Authorization: Bearer $(</var/run/secrets/kubernetes.io/serviceaccount/token)" "$node_url" | jq --unbuffered -c '.object | select(.spec.podCIDR != null)'))
+  [[ -n "${response}" ]] && break
+done
+if [[ -z "${response}" ]]; then
+  fatal "Could not successfully watch node and wait for podCIDR."
+fi
+log "Node object fetched from $node_url"
+log "${response}"
```
marqc, line 134: This whole block may be extracted into the bash function for better readability, i.e. get_k8s_node_descriptor. And additionally the variable name can be changed to something more meaningful than `$response`

marqc (changes requested):

scripts/install-cni.sh
marqc: Additionally you can add the `limit=1` parameter to end the call as soon as the first update on the node object gets returned.

scripts/install-cni.sh
```diff
@@ -110,15 +115,20 @@ else
...
-token=$(</var/run/secrets/kubernetes.io/serviceaccount/token)
 host=${KUBERNETES_SERVICE_HOST}
 # If host contains a colon (:), it is an IPv6 address, hence needs wrapping
 # with [..].
 if [[ "${host}" =~ : ]]; then
   host="[$host]"
 fi
-node_url="https://$host:${KUBERNETES_SERVICE_PORT}/api/v1/nodes/${HOSTNAME}"
-response=$(curl -k -s -H "Authorization: Bearer $token" "$node_url")
+node_url="https://$host:${KUBERNETES_SERVICE_PORT}/api/v1/watch/nodes/${HOSTNAME}"
```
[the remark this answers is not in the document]
jingyuanliang: Done

scripts/install-cni.sh
jingyuanliang: Timeout added, and no, we can't end as soon as the first update, in case the first update is not adding podCIDR.

`grep -m1 . <(...)` can return after receiving the first line.

scripts/install-cni.sh
jingyuanliang: Done.

scripts/install-cni.sh
jingyuanliang: The response is somehow used throughout the file... don't bother to change it for now.

jingyuanliang: Actually, it's not too many. Let me change it and get the opportunity to change and poor styles along the way.

scripts/install-cni.sh
```diff
@@ -110,21 +115,43 @@ else
...
+    [[ -n "${node_object}" ]] && break
+  done
+}
+
+# Watch for up to 1 minute, we don't expect podCIDR to be not populated for too
+# long, but this can also be three continuous retries and failures, then we wait
+# for kubelet to retry the whole container if response is still not fetched.
+fetch_node_object 3 20
+
+if [[ -z "${node_object:-}" ]]; then
```
marqc, line 145: this check and the following 2 log entries can be part of the `fetch_node_object` function as well

jingyuanliang: Modified a bit, but I still like to leave the full dump outside.

marqc (approved):

PR #369, opened 2024-10-07
Use NODE_NAME instead of HOSTNAME for node name if set

jingyuanliang:
... in apiserver query for podCIDR; otherwise fall back to HOSTNAME.

We see unexpected customization of host name on node by users.

/assign @MrHohn


MrHohn (approved):
/lgtm

MrHohn (approved):
/lgtm

PR #370, opened 2024-10-10
pod_ip_metrics: use CURRENT_NODE_NAME env var as the node name if set

jingyuanliang:
and only use os.Hostname() as a fallback to match the previous behavior.

The name CURRENT_NODE_NAME was chosen to match the existing usage in netlink_metrics.

/assign @MrHohn


MrHohn (approved):
Thanks Jingyuan!
/lgtm

jingyuanliang:
cc @basantsa1989

PR #371, opened 2024-10-10
install-cni: rename env var NODE_NAME to CURRENT_NODE_NAME

jingyuanliang:
This is to match the existing usage in netlink_metrics. The version with NODE_NAME naming was never released.

/assign @MrHohn


MrHohn (approved):
/lgtm

PR #373, opened 2024-10-14
Redirect curl errors during node object fetching to stdout

### LINKED ISSUE #93: install-cni.sh should use cacert instead of --insecure
https://github.com/GoogleCloudPlatform/netd/blob/6cd97fdcb3290114cacc2023132bbee7753f9bb5/scripts/install-cni.sh#L75-L79

we are using `curl -k` which is insecure. We should use `--cacert` with the ca cert in the same folder as `token` file.

jingyuanliang:
It appears some "expected errors" still pop up causing user confusion, so do the redirection, restore -m as some further safeguard, and since we're touching this line, fix #93 along the way.

/assign @MrHohn 
/cc @anfernee

### DIFF
--- a/scripts/install-cni.sh
+++ b/scripts/install-cni.sh
@@ -103,6 +103,7 @@ if [[ -z "${cni_spec}" ]]; then
   success
 fi
 
+exec {stdout_dup}>&1
 fetch_node_object() {
   local attempts=$1
   local timeout=$2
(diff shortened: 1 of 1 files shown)


jingyuanliang:
/hold

not tested yet.

jingyuanliang:
/hold cancel

Tested now.

scripts/install-cni.sh
```diff
@@ -103,6 +103,7 @@ if [[ -z "${cni_spec}" ]]; then
   success
 fi
 
+exec {stdout_dup}>&1
```
MrHohn, line 106: Interesting trick.

PR #407, opened 2025-02-12
Always populate ip6tables for cloud.google.com/gke-dpv2-unified-cni

jingyuanliang:
All GKE nodes should have IPv6 stack installed even if the underlying network is not configured for IPv6. Setting the rules for IPv4 single stack clusters should be harmless. Furthermore, this will be gated by cloud.google.com/gke-dpv2-unified-cni (node label) experiment.

/assign @MrHohn 
/cc @sugangli @aojea


aojea:
👍

MrHohn (approved):
/lgtm
/approve

PR #450, opened 2025-07-24
Update TCPInfo Struct 

MaitreyPrajapati:
Current version of netd expects `TCPInfo` struct size to be 232 Bytes. In Kernel version 6.2-rc1+ following additional fields were introduced in [tcp_info](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/tcp.h#n229) struct. 

- Rcv_wnd (uint32 - 4 Bytes)
- Rehash (uint32 - 4 Bytes)
- Total_rto (uint16 - 2 Bytes)
- Total_rto_recoveries (uint16 - 2Bytes)
- Total_rto_time (uint32 - 4Bytes)

When the pod watch is enabled in netd metric, lack of these fields causes excessive noise in logs due to [this](https://github.com/GoogleCloudPlatform/netd/blob/master/pkg/tcp_metrics/parser/parse.go#L241).

This PR updates the LinuxTCPInfo struct to reflect the current version of kernel's tcp_info struct and adds test to verify the struct size against kernel's tcp_info struct to remove [this](https://github.com/GoogleCloudPlatform/netd/blob/master/pkg/tcp_metrics/tcp/tcpinfo_test.go#L9) TODO.

### DIFF
--- a/pkg/tcp_metrics/tcp/tcpinfo.go
+++ b/pkg/tcp_metrics/tcp/tcpinfo.go
@@ -130,4 +130,11 @@ type LinuxTCPInfo struct {
 	RcvOooPack uint32
 
 	SndWnd uint32
+	RcvWnd uint32
+
+	Rehash uint32
+
+	TotalRto           uint16
+	TotalRtoRecoveries uint16
+	TotalRtoTime       uint32
 }
--- a/pkg/tcp_metrics/tcp/tcpinfo_test.go
+++ b/pkg/tcp_metrics/tcp/tcpinfo_test.go
@@ -2,11 +2,22 @@ package tcp_test
 
 import (
 	"testing"
+	"unsafe"
 
 	"github.com/GoogleCloudPlatform/netd/pkg/tcp_metrics/tcp"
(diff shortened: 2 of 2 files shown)


pkg/tcp_metrics/tcp/tcpinfo.go
```diff
@@ -130,4 +130,10 @@ type LinuxTCPInfo struct {
 	RcvOooPack uint32
 
 	SndWnd uint32
+	RcvWnd uint32
+
+	Rehash uint32
+
+	TotalRto           uint16
+	TotalRtoRecoveries uint32
```
MrHohn, line 138: One quick question - from the latest `tcp_info` struct, there is also a `tcpi_total_rto_time ` field, curious why that isn't included?

MrHohn (commented):
Thanks @MaitreyPrajapati! The change generally looks good to me.

Could you amend the commit message to be more contextual? Or it may look a bit confusing purely from the git history.

MrHohn (approved):
/lgtm
/approve

PR #452, opened 2025-07-28
Fix the initialization for policy routing to fatal on errors

MrHohn:
and add a retry logic for retrieving the node.

The init func for config package is adjusted to be a usual func, otherwise it won't work in a unit test environment.

/assign @jingyuanliang


MrHohn:
Local verification done for the changes now. PTAL, thanks!

pkg/config/config_policy_routing.go
```diff
@@ -89,7 +91,9 @@ var PolicyRoutingConfigSet = Set{
 	nil,
 }
 
-func init() {
+// InitPolicyRouting performs necessary initialization for policy routing.
+// It should be called before running the policy routing enforcement loop.
+func InitPolicyRouting() {
```
jingyuanliang, line 96: Since it's no longer the special init() function, it can return an error. I feel it's better to return the error all the way and Fatal() in main(). Alternatively, the functions in this calling chain should be called MustXXX() to indicate that it can crash the whole program (while I prefer returning errors).

pkg/config/config_policy_routing.go
```diff
@@ -254,13 +255,25 @@ func init() {
...
 	// Retrieve necessary IP info from the node object.
-	node, err := clientset.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("getting node %s: %w", nodeName, err)
+	var node *v1.Node
+	if err := wait.PollUntilContextTimeout(context.Background(), 2*time.Second, 30*time.Second, true, func(context.Context) (bool, error) {
+		var err error
+		node, err = clientset.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
+		if err != nil {
+			glog.Errorf("getting node %s: %w", nodeName, err)
```
jingyuanliang, line 271: https://google.github.io/styleguide/go/decisions#error-strings

And I think glog.Errorf doesn't take %w?

pkg/config/config_policy_routing.go
```diff
@@ -254,13 +255,25 @@ func init() {
...
 
 func fillLocalRulesFromNode(clientset kubernetes.Interface, nodeName string) error {
 	// Retrieve necessary IP info from the node object.
-	node, err := clientset.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("getting node %s: %w", nodeName, err)
+	var node *v1.Node
+	if err := wait.PollUntilContextTimeout(context.Background(), 2*time.Second, 30*time.Second, true, func(context.Context) (bool, error) {
+		var err error
+		node, err = clientset.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
```
jingyuanliang, line 269: use the ctx passed in from func arg?

pkg/config/config_policy_routing.go
MrHohn: Thanks - fixed now.

pkg/config/config_policy_routing.go
MrHohn: Make sense  - done.

pkg/config/config_policy_routing.go
```diff
@@ -254,13 +255,25 @@ func init() {
...
 }
 
-func fillLocalRulesFromNode(clientset kubernetes.Interface, nodeName string) error {
+func fillLocalRulesFromNode(ctx context.Context, clientset kubernetes.Interface, nodeName string) error {
 	// Retrieve necessary IP info from the node object.
-	node, err := clientset.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("getting node %s: %w", nodeName, err)
+	var node *v1.Node
+	if err := wait.PollUntilContextTimeout(ctx, 2*time.Second, 30*time.Second, true, func(context.Context) (bool, error) {
```
jingyuanliang, line 267: `func(ctx context.Context) (bool, error) {` ?

MrHohn: Fixed now.

PR #455, opened 2025-08-09
Prevent InitPolicyRouting from resetting configs

MrHohn:
Previously, calling `InitPolicyRouting` would reset the global `PolicyRoutingConfigSet.Configs` slice. This could lead to the loss of configurations if the initialization order was not strictly controlled or if the function were ever called multiple times.

This commit corrects the behavior by refactoring out the global variable for policyRoutingConfigSet.

To properly test this fix and improve the overall design, `InitPolicyRouting` has been refactored to accept its dependencies (Kubernetes clientset, node name) as arguments. This enables dependency injection and isolated unit testing.

A new few new unit tests have been added to verify the corrected behavior and ensure that configurations are appended as expected.

/assign @jingyuanliang

### DIFF
--- a/pkg/config/config_policy_routing.go
+++ b/pkg/config/config_policy_routing.go
@@ -34,7 +34,6 @@ import (
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
 
-	"github.com/GoogleCloudPlatform/netd/pkg/utils/clients"
 	"github.com/GoogleCloudPlatform/netd/pkg/utils/nodeinfo"
 )
 
@@ -84,6 +83,11 @@ var (
 	vethGatewayDst   net.IPNet
 )
 
+var (
+	RouteGet    = netlink.RouteGet
+	LinkByIndex = netlink.LinkByIndex
+)
+
 // PolicyRoutingConfigSet defines the Policy Routing rules
 var PolicyRoutingConfigSet = Set{
 	false,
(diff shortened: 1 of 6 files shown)


pkg/config/config_policy_routing.go
```diff
@@ -88,22 +87,27 @@ var (
 var PolicyRoutingConfigSet = Set{
 	false,
 	"PolicyRouting",
-	nil,
+	[]Config{},
```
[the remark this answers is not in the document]
MrHohn: Nope - I was under the wrong impression we may need an empty slice to append to. Turned out I was wrong. Fixed now.

MrHohn:
@jingyuanliang I have made a few more changes to refactor out the global variable dependency.

pkg/config/config_policy_routing.go
```diff
@@ -121,21 +121,13 @@ func InitPolicyRouting(ctx context.Context) error {
...
 
-	clientset, err := clients.NewClientSet()
-	if err != nil {
-		return err
-	}
-	nodeName, err := nodeinfo.GetNodeName()
-	if err != nil {
-		return err
-	}
 	if err := fillLocalRulesFromNode(ctx, clientset, nodeName); err != nil {
```
jingyuanliang, line 124: This one is another global state usage...

pkg/config/config_policy_routing.go
```diff
@@ -254,11 +246,12 @@ func InitPolicyRouting(ctx context.Context) error {
 			RuleDel:  netlink.RuleDel,
 			RuleList: netlink.RuleList,
 		},
-	}
+	}...)
+
 	glog.Info("Including local table rules.")
-	PolicyRoutingConfigSet.Configs = append(PolicyRoutingConfigSet.Configs, LocalTableRuleConfigs...)
+	policyRoutingConfigSet.Configs = append(policyRoutingConfigSet.Configs, LocalTableRuleConfigs...)
```
jingyuanliang, line 252: (then used the global state here)

PR #462, opened 2025-08-26
feat(netconf): Stop installing UDP mitigation rule for newer kernels

MrHohn:
The workaround for a kernel issue causing problems with local UDP traffic is only necessary for kernel versions 6.6.57 up to (but not including) 6.6.94.

This commit updates the logic to avoid installing the mitigation IP rule on newer kernels where the issue has been resolved.

The corresponding unit tests have been updated to verify the new behavior.

Reference: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=50d9ce9679dd50df2dc51ada717fa875bc248fad

/assign @jingyuanliang


pkg/controllers/netconf/network_config_controller.go
```diff
@@ -62,8 +63,11 @@ func NewNetworkConfigController(enablePolicyRouting, enableSourceValidMark, excl
 		glog.Errorf("Could not check kernel version: %v. Skip installing UDP exempt rule.", err)
 	} else {
 		glog.Infof("Kernel version detected: %v.", kernelVersion)
-		if kernelVersion.AtLeast(version.MustParseGeneric(brokenLocalUDPKernelVersionStart)) {
-			glog.Infof("Kernel version is impacted by a known issue (start version: %v). Including an IP rule to exempt UDP traffic.", brokenLocalUDPKernelVersionStart)
+		// The known issue was fixed since linux kernel 6.6.94.
+		if kernelVersion.AtLeast(version.MustParseGeneric(brokenLocalUDPKernelVersionStart)) &&
+			!kernelVersion.AtLeast(version.MustParseGeneric(brokenLocalUDPKernelVersionEnd)) {
```
jingyuanliang, line 68: https://google.github.io/styleguide/go/decisions#indentation-confusion

MrHohn:
Converted to draft as I realized we may need to handle for version higher than 6.6.X.

pkg/controllers/netconf/network_config_controller.go
MrHohn: Done.

MrHohn:
Okay, converted to version ranges instead as per the discussion yesterday. PTAL thanks!

PR #472, opened 2025-09-24
Avoid depending on httpbin.org service

jingyuanliang:
Use https://www.google.com/generate_204 instead.

/assign @MrHohn


MrHohn (approved):
/lgtm

PR #522, opened 2026-06-23
Fix file descriptor leaks of netns handles and sockets in netlink metrics collector.

MrHohn:
- Add defer basens.Close() in getSnapshots to release base network namespace file descriptor.
- Defer a single cleanup function at the beginning of getSnapshots to close all open namespace handles in the namespaces slice on function exit.
- Explicitly close the netlink socket `s` on every return path inside getSnapshots loop, and at the end of the iteration, removing the unsafe defer-inside-loop pattern.
- Close `ns` handle in getNamespaces when unix.Fstat fails.


jingyuanliang:
@MrHohn please fix lint errors

MrHohn:
> @MrHohn please fix lint errors

Done

